PR #3, opened 2017-07-25
Buck build support

njlr:
This PR adds support for builds using [Buck](https://buckbuild.com/).

To build the library:

```bash=
buck build :monocypher
```

The existing Make build is unchanged; the two can coexist peacefully 😊

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -1,0 +1,2 @@
+[project]
+  ignore = .git
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,10 @@ sodium
 donna
 speed
 rename_*
+
+# Buck
+/buck-out/
+/.buckd/
+/buckaroo/
+.buckconfig.local
+BUCKAROO_DEPS
--- a/BUCK
+++ b/BUCK
@@ -1,0 +1,12 @@
+cxx_library(
+  name = 'monocypher',
+  exported_headers = subdir_glob([
+    ('src', '**/*.h'),
+  ]),
+  srcs = glob([
+    'src/**/*.c',
+  ]),
+  visibility = [
+    'PUBLIC',
+  ],
+)
--- a/README.md
+++ b/README.md
@@ -19,6 +19,10 @@ just copy `src/monocypher.h` and `src/monocypher.c` into your project.
 They compile as C99, C11, C++98, C++11, C++14, and C++17. (Tested with
 gcc 5.4.0 and clang 2.8.0 on GNU/Linux.)
 
+Or, using [Buckaroo](https://buckaroo.pm):
+
+    $ buckaroo install github+loupvaillant/monocypher
+
 Test suite
 ----------
 
--- a/buckaroo.json
+++ b/buckaroo.json
@@ -1,0 +1,4 @@
+{
+  "name": "Monocypher",
+  "target": "monocypher"
+}


LoupVaillant:
Hmm, I'm conflicted. I don't like to turn your patch down, but I think another approach is warranted.

See, Monocypher doesn't have any build system. They're out of scope. The makefile is there to _test_ Monocypher, not to build or install it.

On the other hand, having alternate _distributions_ of Monocypher would be very nice. If someone makes a Rust binding, it should be on Cargo. If someone wants to put it in a Linux distribution, it should have its `.deb` or `.rpm` package. Your patch should be used, _somehow_.

I also have 2 minor qualms about your patch: first, it seems to use C++ only, I don't see any option for C. Second, I'm not sure the user can compile Monocypher with SHA-512 support (needed for ed25519) but that's less important, since it is clearly not the default.

Here's what I think you should do:

1. Fork Monocypher.
2. Remove everything, except the `src` directory.
    -If you want to stick with the default settings, also remove `sha512.c` and `sha512.h`. If you want to allow the non-default, but more official ed25519 (some users will want it), keep them, and allow that option in the build system (I have macros that activate when you compile with -DED25519_SHA512)
3. Add your build system.
4. Rewrite your Readme as you see fit. A link back to my own repo or to my [website](http://loup-vaillant.fr/projects/monocypher/)) would be very nice.
5. Send another pull request to modify _my_ Readme, just like you did now (except this time you will point to your distribution).

I think forks are scary because of the synchronisation burden: upstream changes need to propagate. Monocypher however is expected to be _extremely_ stable. With any luck, the contents of the `src` directory will never have to change again, and neither will forks, distributions, and language bindings.

What do you think? Is my approach sound?

njlr:
> See, Monocypher doesn't have any build system. They're out of scope. The makefile is there to test Monocypher, not to build or install it.

Buck is a containerization system and a build system. Buck provides a way to describe a header-only library such that you can depend on it and Buck will generate the appropriate include paths for Clang / GCC. 

> I also have 2 minor qualms about your patch: first, it seems to use C++ only, I don't see any option for C. Second, I'm not sure the user can compile Monocypher with SHA-512 support (needed for ed25519) but that's less important, since it is clearly not the default.

Buck allows the language standard to be decided on a per-target basis or globally. Globally is preferrable, since it ensures a consistent ABI across targets. I have not set the language standard in my build script so that the user can decide globally. If they choose a C langauge standard it will build in C mode. 

The same mechanism can be used for flags like `-DED25519_SHA512` 😊

Your solution sounds good. Actually, I might be able to do something with Submodules, where the Buck script points to code inside of a submodule. I'll keep you posted!

LoupVaillant:
Excellent, thank you. I see my qualms were unjustified, that's good news. :-)

I'll wait for your modifications before we merge this.

CuleX:
@njlr It's been a while, can I ask if you're still working on this?

njlr:
I'm afraid that other priorities have come up. It would be great if you chose to merge this, but for now I am just redirecting people to my repo: 

```
buckaroo install github+njlr/monocypher
```

LoupVaillant:
That's neat! I didn't know you did that, we need to update the README (it now has a list of known alternative distributions). Do you want to propose a patch for the README (and plant your flag), or do you rather had me do it?

Once this is done, I think we should close this pull request, if that's okay with you. Buck build support is neat, but I don't see a way to integrate it to the main repository, especially considering the cleanup I have in mind. And then even if I can integrate Buck, I'll have to consider other build systems. I sense scope creeping in…

njlr:
I have added a note to the README about my Buck port in this PR: https://github.com/LoupVaillant/Monocypher/pull/26

Please feel free to change / approve / reject as you like!

PR #13, opened 2017-08-06
Test constant-timeness of comparison functions

CuleX:
clock() in time.h is specified in C89 and should thus be portable.
COMPARISON_DIFF_THRESHOLD was chosen based on observed values, which
reached a maximum at 3.

### DIFF
--- a/tests/self.c
+++ b/tests/self.c
@@ -1,6 +1,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #include "monocypher.h"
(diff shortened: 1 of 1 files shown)


CuleX:
I know you're going to some great extents to prevent compilers from optimizing the comparison into a regular memcmp or similar, but today's smartness might not best tomorrow's compiler. I figure it's best to check whether the safety of comparison really holds true.

LoupVaillant:
Excellent idea, thank you. As far as portability is concerned, it would seem even the TIS-interpreter can deal with this.

PR #28, opened 2017-09-07
Minor improvements to crypto_key_exchange man page

CuleX:
1. Change param "shared_secret" in crypto_x25519 to "raw_shared_secret" to aid quick identification of the difference between it and crypto_key_exchange; the "key" vs. "secret" gets lost easily because of the shared prefix "shared_".
2. Fix not having punctuation as a separate word in macro.

### DIFF
--- a/man/3monocypher/crypto_key_exchange.3monocypher
+++ b/man/3monocypher/crypto_key_exchange.3monocypher
@@ -21,7 +21,7 @@
 .Fc
 .Ft int
 .Fo crypto_x25519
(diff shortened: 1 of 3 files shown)


LoupVaillant:
Good idea. I think we should modify the source code as well —ideally in the same commit. Could you do it?

CuleX:
This look good? The tests continue passing.

LoupVaillant:
Perfect. Merged, thanks.

PR #30, opened 2017-09-09
Man page fixes

CuleX:
Various things that I noticed while trying to use Monocypher.


LoupVaillant:
Excellent work. Merging right away!

PR #31, opened 2017-09-13
crypto_sign man page improvements

CuleX:


LoupVaillant:
Good catch. Do you think it would be worthwhile to use a local array for the signature, so it *can* overlap with the message? Or should we just go ahead and `restrict` them?

Thinking about it, I may have to review the rest of the library for this kind of issue…

CuleX:
> Good catch. Do you think it would be worthwhile to use a local array for the signature, so it can overlap with the message?

Monocypher is a library. I don't particularly enjoy libraries putting (avoidable) restrictions on my coding, personally. An additional 64 bytes with implicit automatic storage duration, given everything else, don't sound very severe (`h_ram` already uses 64 bytes, `s` later on also does) even assuming you target 32-bit embedded platforms.

I didn't check, but I think you wouldn't need a new local array, you could just repurpose `signature` twice, once for `ge_tobytes` and once for `hash_ram`; obviously, if either the SHA-512 or the BLAKE2b implementation don't allow for overlapping buffers, the issue just shifts up a bit.

> I may have to review the rest of the library for this kind of issue…

That's probably a good idea.

LoupVaillant:
Okay, I'll lift the restriction

PR #40, opened 2017-09-23
Create all installation directories

CuleX:
This is required for the `DESTDIR` variable to actually work and create the necessary tree.


CuleX:
Maybe some context on `DESTDIR`: People use it to capture the full installation for a package.

A simple example from a PKGBUILD: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=alpine#n50

LoupVaillant:
Good catch. Now that I think of it, I think we're still missing the pkg-config directory. I'll take care of it.

PR #51, opened 2017-10-04
Fix mandoc invocation for recent mandoc versions

CuleX:
A commit in mandoc earlier this year subtly broke the `-O` parsing.
Multiple instances of `-O` do not get parsed, so all options have to be passed into the same `-O` with comma separation as intended.

Presumably caused by [this commit](https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/mandoc/main.c.diff?r1=1.185&r2=1.186&f=h).


LoupVaillant:
It works, thanks!

PR #52, opened 2017-10-06
Fix formatting and typos in the man pages; note Ed25519 fault injection on crypto_sign

CuleX:


CuleX:
I took the liberty to make a design choice with documenting the Ed25519 fault injection note. Rationale:

1. The protection by a provided by a verification pass is neither absolute nor fast; the same applies to multiple computation (p. 5)
2. Monocypher vehemently refuses to write an operating system detection to provide a RNG function that could be used for a random R value (as suggested as mitigation on p. 5). This has good reasons that are not worth overturning.
3. Mixing hardware and software implementations (p. 5) is not an actual option in the context Monocypher finds use.

I thus skipped to the conclusion of *caveat emptor*. Feel free to reverse 0946d11 if you disagree.

LoupVaillant:
I see you read my comment. :smile: 

Your patch is perfect, let's merge it all.

PR #59, opened 2017-11-21
crypto_lock manual tweaks

mikejsavage:
Not entirely sure the bit about ad length is an improvement!

### DIFF
--- a/doc/man/man3/crypto_lock.3monocypher
+++ b/doc/man/man3/crypto_lock.3monocypher
@@ -149,8 +149,10 @@ and
 .It Fa ad_size
 length of the additional data.
 .Sy That length is not authenticated.
-If it is transmitted over the wire, it should be appended to the
-additional data, and authenticated with it.
+If the additional data is of variable length, the length should be appended to
+.Fa ad
+so it gets authenticated, and should be extracted from the end of the
+message when decrypting.
 Otherwise an attacker could provide a false length, effectively moving
 the boundary between the additional data and the ciphertext.
 This may cause buffer overflows in some programs.
@@ -177,6 +179,8 @@ and
 .Fa cipher_text ) .
 Corruption can happen because of transmission errors, programmer
 error, or an attacker's interference.
+.Fa plain_text
+does not need wiping if the decryption fails.
 .Sh EXAMPLES
 Encryption:
 .Bd -literal -offset indent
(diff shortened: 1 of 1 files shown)


doc/man/man3/crypto_lock.3monocypher
```diff
@@ -149,8 +149,10 @@ and
 .It Fa ad_size
 length of the additional data.
 .Sy That length is not authenticated.
-If it is transmitted over the wire, it should be appended to the
-additional data, and authenticated with it.
+If the additional data is of variable length, the length should be appended to
+.Fa ad
+so it gets authenticated, and should be extracted from the end of the
```
CuleX: Maybe "so that"? It feels like that would flow better, but YMMV.

doc/man/man3/crypto_lock.3monocypher
```diff
@@ -177,6 +179,8 @@ and
 .Fa cipher_text ) .
 Corruption can happen because of transmission errors, programmer
 error, or an attacker's interference.
+.Fa plain_text
+does not need wiping if the decryption fails.
```
CuleX: *need to be wiped

doc/man/man3/crypto_lock.3monocypher
```diff
@@ -218,8 +220,10 @@ size_t         text_size;    /* Message size (NOT secret) */
 uint8_t       *plain_text;   /* Secret message            */
 if (crypto_unlock(plain_text, key, nonce, mac,
                   cipher_text, text_size)) {
-    /* The message is corrupted */
-    /* Abort the decryption     */
+    /* The message is corrupted           */
+    /* Wipe key if it is no longer needed */
+    /* and abort the decryption           */
```
CuleX: I know I'm late since the original did it, but shouldn't we be doing this instead?

    /* The message is corrupted.
     * Wipe key if it is no longer needed
     * and abort the decryption.
     */

Applies to the next one, too.

doc/man/man3/crypto_lock.3monocypher
```diff
@@ -94,11 +94,9 @@ The secret message.
 Its content will be kept hidden from attackers.
 Its length however, will
 .Em not .
-Make sure message lengths do not leak secret information.
-This has happened before with variable-length voice
-encoding software.
-Solutions to mitigate this include constant-length encodings and
-padding.
+Be careful when combining encryption with compression. See
```
CuleX: Newline after period.

doc/man/man3/crypto_lock.3monocypher
LoupVaillant: My mileage would keep "that" out. :smile:

doc/man/man3/crypto_lock.3monocypher
LoupVaillant: Correct, this is lighter that way. I should have done it like this from the start.

PR #60, opened 2017-11-21
Argon manual tweaks

mikejsavage:


CuleX:
Looks good to me.

doc/man/man3/crypto_argon2i.3monocypher
```diff
@@ -64,7 +64,8 @@ Must be
 .Fa nb_blocks
 × 1024 bytes big, and suitably aligned for 64-bit integers.
 .Fn malloc
-should yield a suitable chunk of memory.
+will yield a suitable chunk of memory on platforms that support 64-bit types
+(e.g. int64_t or IEEE double precision floats).
```
LoupVaillant: Actually, it will work on *any* platform where Monocypher compiles at all. I believe this is written explicitly in the standard. For instance, on 32-bit platforms, `malloc()` may return a 4-byte aligned chunk of memory, but that's still enough to handle 64-bit integers: since they're represented by 2 32-bit words, we only need 32-bit alignment.

`will yield a suitable chunk of memory.` is enough.

doc/man/man3/crypto_argon2i.3monocypher
```diff
@@ -141,22 +142,22 @@ Use
...
+decided how long the computation should take.
+For user authentication, somewhere between half a second (convenient)
+and several seconds (paranoid) is a good starting point.
+The computation should use as much memory as can be spared.
+.Pp
+Since parameter selection depends on your hardware, some trial and error
+will be required in order to determine the ideal settings.
+Three iterations and 100000 blocks (that is, one hundred megabytes of
+memory) is a good starting point.
 Adjust
```
LoupVaillant: Good change. I would remove the first instance of "is a good starting point", though: the time it takes to hash the password is not a starting point, it's the goal; and I like to avoid repetitions

doc/man/man3/crypto_argon2i.3monocypher
```diff
@@ -172,7 +173,7 @@ const uint8_t  salt[16];                    /* random salt     */
 const uint32_t nb_blocks = 100000;          /* 100 megabytes   */
 const uint32_t nb_iterations = 3;           /* 3 iterations    */
 void *work_area = malloc(nb_blocks * 1024); /* work area       */
-if (work_area == 0) {
+if (work_area == NULL) {
```
LoupVaillant: Good catch. I say `0` all over my code, but `NULL` is indeed better for the manual.

PR #61, opened 2017-11-21
Explicitly talk about compression being harmful in the intro

mikejsavage:


mikejsavage:
@CuleX FYI

doc/man/man3/intro.3monocypher
```diff
@@ -177,14 +177,13 @@ Some older processors do not multiply in constant time.
...
 .Pp
-.Sy The lengths of the inputs are not secret.
-Timings do reveal them \(en So do network traffic and file sizes.
-Most of the time, lengths do not contain enough information for this
-to be a problem.
-Sometimes however they do.
-It has happened before with variable-length voice encoding software.
-The researchers managed to identify the speakers and recover parts of
-the conversation.
+.Ss Data compression
```
CuleX: Can we maybe merge this into something more generic?

e.g. It's not *only* compression that can be harmful. Imagine you're in some kind of chat protocol and you're trying to put a nickname (whose IP address you don't have) to a person (whose network you're hooked into). Message lengths would give that away really fast. Your point about encryption still stands, though.

LoupVaillant: I confess I can't find a better title for this sub section.

The one thing I would like to see preserved is the mention that even passive attacks can work in some cases. But I'm not sure how to do it without cluttering the text too much, and the CRIME example is scarier and  better anyway.

It also says what I care about right after, with lengths not being secrets. And while I mourn the removal of my example a little, the replacement is better.

I may have one qualm: there's no mention that even _passive_ attacks may work.

PR #62, opened 2017-11-26
Add symlink for crypto_argon2i_general.3monocypher

CuleX:
This also adds it to intro(3monocypher).


LoupVaillant:
Good catch, thanks.

PR #65, opened 2017-11-28
BLAKE2b manual tweaks

mikejsavage:


doc/man/man3/crypto_blake2b.3monocypher
```diff
@@ -98,18 +94,19 @@ Must be between 0 and 64.
...
+.Fn crypto_blake2b
 and
-.Fn crypto_blake2b ,
-provided for convenience (calling it is the same as calling
+.Fn crypto_blake2b_general .
+.Fn crypto_blake2b
+is provided for convenience, and is equivalent to calling
 .Fn crypto_blake2b_general
-with a null key and a 64-byte hash).
+with no key and a 64 bit hash.
```
CuleX: No, it really is a 64-*byte* hash. A 64-bit hash would be eight bytes, which is far too little.

mikejsavage: doh, fixed

PR #69, opened 2017-11-28
crypto_verify manual tweaks

mikejsavage:


doc/man/man3/crypto_verify16.3monocypher
```diff
@@ -17,16 +17,15 @@
 .Sh DESCRIPTION
 Cryptographic operations often require comparison of secrets or values
 derived from secrets.
-Standard comparison functions tend to exit as soo as they find a
-difference, and doing so leaks information through variable timings.
+Standard comparison functions (like memcmp) tend to exit when they find
```
CuleX: Maybe we should remove the parentheses here?

doc/man/man3/crypto_verify16.3monocypher
```diff
@@ -17,16 +17,15 @@
...
-A MAC may be sent over the network along with a message.
-However, the correct MAC is secret.
+As an example, say a message authentication code (MAC) is sent over the
+network along with a message, but the correct MAC is secret.
 If the attacker attempts a forgery, one does not want to reveal
 .Dq your MAC is wrong, Em and it took 384 microseconds to tell .
-If the next attempt takes 462 microseconds instead, it gives away the
-fact that the attacker just guessed a few bytes correctly.
+If the next attempt takes 462 microseconds instead, it tells the
+attacker they just guessed a few bytes correctly.
```
CuleX: I'd think "that" would improve readability, but YMMV.

doc/man/man3/crypto_verify16.3monocypher
mikejsavage: agreed, matches the "like memset and bzero" in the crypto_wipe PR

doc/man/man3/crypto_verify16.3monocypher
mikejsavage: I am fine either way, let's see what Loup thinks

LoupVaillant: I think I prefer it as is. I don't like to add redundant "that" when it can be avoided (yes, I did even worse just there).

Issue #29, opened 2017-09-08
Incremental crypto_lock/crypto_sign API


CuleX:
It may be worthwhile to also create an incremental `crypto_lock`/`crypto_unlock` API for cases where you have multiple buffers to decrypt (e.g. header and body in separate buffers) or a large file to read out in chunks.

Same for `crypto_sign`/`crypto_check` (think: signing fairly large update files).

YMMV, but it's an idea.

CuleX:
# crypto_lock

> `crypto_unlock` highlights a dilemma: a one pass incremental interface has to interleave authentication and decryption. If the message is corrupted, you lost 3 times more time than you would have if you only did the authentication.

Isn't that a tradeoff you can shift to the user instead? I'm considering the following situations:

* Multiple medium-sized buffers in memory where allocating one for the whole message may risk violating memory constraints. If you're dealing with corrupted memory, the lost time is the least of your issues.
* One very large file read in chunks (where `mmap()` is not an option e.g. because you're on a 32-bit platform without PAE and the file's beyond what you could map in a 32-bit address space, or because you're looking at an embedded platform whose kernel *can't* `mmap()`), primarily from at-rest AE. The only failure methods I can think of are an attacker (who's clearly got no sane plan) and random corruption. The latter is, again, quite a lot more grave than the other, dwarving the performance of your call. The former is, depending on your threat model, unlikely enough that you can sacrifice the performance for the memory cost improvements.
* Multiple individual small messages on the network. *This* is where your performance loss becomes frustrating. I believe TCP can genreally be assumed to be reliable these days and corruption is fairly likely an attempt at an attack. There's a tradeoff depending on the threat model since you're probably going to shortcut a failure here, and that's IMHO best decided by the user, rather than the library.

---

# crypto_sign

## Solution involving documentation

> The situation is somewhat reversed for signing. Signature verification only requires one pass, so an incremental interface is easy. To sign a message however, you have to hash it twice. 

Okay, who thought this was a good idea?

[...]

PR #70, opened 2017-12-06
Incremental crypto_lock manual tweaks

mikejsavage:


doc/man/man3/crypto_lock_init.3monocypher
```diff
@@ -64,43 +64,42 @@ and
...
-.Fn crypto_unlock_final
-to verify the MAC.
+Generate the MAC with
+.Fn crypto_lock_final .
 .El
 .Pp
-Because the incremental interface interleaves encryption and
-authentication, it treats corrupted messages three times slower than
-the direct interface.
-Users who expect a high corruption rate may want a different approach.
```
mikejsavage: If you think this should be kept in I can put it back and rephrase it a little

CuleX: Yes, this should be kept. It was quite a long struggle to get it in because [the incremental interface itself was quite controversial](https://github.com/LoupVaillant/Monocypher/issues/29) precisely *because* of this speed tradeoff.

mikejsavage: I stuck it in an implementation details block at the end

it doesn't seem super important to me

LoupVaillant: I agree with putting that paragraph in the implementation details section.

By the way, I learned about another, much more important, reason why incremental processing is controversial: the possibility of trusting unauthenticated messages.

As such, incremental authentication (or signature verification for that matter) isn't problematic. What is is the idea of processing *unauthenticated* data further down the processing pipeline. And if the message turned out to be a forgery, such processing must be cleanly undone, and mustn’t have triggered any side effects, perhaps even timed responses that may inform the attacker about the contents of the message.

It's a security nightmare.

Whatever the user does, she must not trust the message before the authentication code has been verified. In practice, this generally means simply dumping the message to disk, and ignoring (or erasing) it if it's a forgery.

I think we should add a "security considerations" section for this.

PR #71, opened 2017-12-06
Poly1305 manual tweaks

mikejsavage:
I didn't make any notes for this one in the big docs feedback issue (must have missed it), so please make sure you agree with my changes

### DIFF
--- a/doc/man/man3/crypto_poly1305.3monocypher
+++ b/doc/man/man3/crypto_poly1305.3monocypher
@@ -33,31 +33,30 @@
 .Fa "const uint8_t mac[16]"
 .Fc
 .Sh DESCRIPTION
-Poly1305 is a one-time message authentication code.
-"One time" means
-the authentication key can be used only once.
-This makes Poly1305
-.Sy easy to mess up .
-On the other hand, Poly1305 is fast, as well as provably secure if
-used correctly.
+Poly1305 is a one-time message authentication code. "One-time" means the
+authentication key can be used only once.
+.Sy This makes Poly1305 easy to misuse .
+On the other hand, Poly1305 is fast, as well as provably secure if used
+correctly.
 .Pp
 Poly1305 is a low-level primitive.
 Consider using authenticated encryption, implemented by
 .Xr crypto_lock 3monocypher .
 .Ss Direct interface
 .Fn crypto_poly1305
-produces a message authentication code for the given
-message and authentication key.
-The authentication key must be used only once.
+produces a message authentication code for the given message and
+authentication key.
+The
 .Fa mac
 and the
 .Fa message
 arguments may overlap.
 .Pp
-Once the authentication is done, the
+The
 .Fa key
 should be wiped with
-.Xr crypto_wipe 3monocypher .
+.Xr crypto_wipe 3monocypher
+after being used.
 .Pp
 To verify the integrity of a message, use
 .Xr crypto_verify16 3monocypher
@@ -66,7 +65,6 @@ to compare the received MAC to the output
 .Ss Streaming interface
 .Fn crypto_poly1305_init
 initialises a context.
-The
 .Fa key
 should be wiped once the context is initialised.
 Then,
@@ -85,7 +83,8 @@ const uint8_t msg[500];  /* Message to authenticate           */
 const uint8_t key[ 32];  /* Random secret key (use only once) */
 uint8_t       mac[ 16];  /* Message authentication code (MAC) */
 crypto_poly1305(mac, msg, 500, key);
-crypto_wipe(key, 32);    /* The key should be wiped after use */
+/* Wipe the key */
+crypto_wipe(key, 32);
 .Ed
 .Pp
 To verify the above message:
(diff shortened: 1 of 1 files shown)


doc/man/man3/crypto_poly1305.3monocypher
```diff
@@ -144,25 +145,22 @@ The attacker must not be able to guess the key;
...
+Session keys cannot be used for this.
+They are shared and secret, but would be reused when sending multiple
+messages.
+Random numbers cannot be used either as there is no reasonable way to
+share it with the recipient without also revealing it to the attacker.
 .Pp
 The only practical source for the authentication key is a chunk of the
 encryption stream used to encrypt the message.
 That chunk must be
 .Em dedicated
```
mikejsavage: I don't understand this paragraph.

Does it go like this:

The "encryption stream" is like a random number stream seeded by the encryption key that you XOR with or whatever to encrypt the data. To authenticate a message you pull a 32 byte Poly1305 key from the head of the stream and do not use those bytes to encrypt any data (keyword in the text being "dedicated"), and then you encrypt your message with the rest of the stream

mikejsavage: I guess that is pretty much what it says in the next paragraph. maybe this one should be taken out

doc/man/man3/crypto_poly1305.3monocypher
```diff
@@ -33,31 +33,30 @@
...
 .Fc
 .Sh DESCRIPTION
-Poly1305 is a one-time message authentication code.
-"One time" means
-the authentication key can be used only once.
-This makes Poly1305
-.Sy easy to mess up .
-On the other hand, Poly1305 is fast, as well as provably secure if
-used correctly.
+Poly1305 is a one-time message authentication code. "One-time" means the
```
CuleX: Newline after period.

doc/man/man3/crypto_poly1305.3monocypher
```diff
@@ -66,7 +65,6 @@ to compare the received MAC to the output
 .Ss Streaming interface
 .Fn crypto_poly1305_init
 initialises a context.
-The
```
CuleX: You removed "The" here, but added it above in a similar context. Why?

doc/man/man3/crypto_poly1305.3monocypher
```diff
@@ -125,17 +126,17 @@ crypto_poly1305_final(&ctx, mac);
 These functions implement Poly1305, described in RFC 7539.
 .Sh  SECURITY CONSIDERATIONS
```
CuleX: A bit late to notice, but double space after macro shouldn't be there.

mikejsavage:
thanks, fixed those

doc/man/man3/crypto_poly1305.3monocypher
LoupVaillant: Yes, that's exactly what it means.

doc/man/man3/crypto_poly1305.3monocypher
```diff
@@ -144,25 +145,22 @@ The attacker must not be able to guess the key;
...
-.Em reused .
-The attacker will recover it and break all security.
-A simple random number cannot be used either: there is no reasonable
-way to give it to the recipient without also revealing it to the
-attacker.
+Session keys cannot be used for this.
+They are shared and secret, but would be reused when sending multiple
+messages.
+Random numbers cannot be used either as there is no reasonable way to
+share it with the recipient without also revealing it to the attacker.
```
LoupVaillant: > share _them_ with the recipient without also revealing _them_ to the attacker.

I think?

doc/man/man3/crypto_poly1305.3monocypher
```diff
@@ -173,23 +171,18 @@ Take the first 32 bits of that stream as your authentication key, then
...
 .Ss Protection against side channels
 Use
 .Xr crypto_verify16 3monocypher
-To compare message authentication code.
-Avoid standard buffer comparison functions.
-They do not run in constant time, which allows attackers to recover
-the MAC in relatively few tries.
+to compare message authentication codes.
+Avoid standard buffer comparison functions, as they may not run in
+constant time.
```
LoupVaillant: I agree with the change, but I wonder if we want to add a little additional warning: the problem with standard comparison functions is that timing attacks are not only possible, they are almost *easy*, even over the network. Other algorithms (Argon2d, naive AES, scrypt…) don't run in constant time, yet remain harder to attack.

Maybe add something like "This allows timing attacks" or something? I know it's redundant, and I don't like the ugliness, but I think this is worth emphasising.

doc/man/man3/crypto_poly1305.3monocypher
mikejsavage: yep

doc/man/man3/crypto_poly1305.3monocypher
mikejsavage: yeah I guess if the manual has warnings it should explain them rather than leaving them as mysteries

PR #72, opened 2017-12-07
Key exchange manual tweaks

mikejsavage:
my most ambitious review yet, this one gives crypto_x25519 a man page to itself... (like poly1305/chacha20/chacha20H)

### DIFF
--- a/doc/man/man3/crypto_x25519.3monocypher
+++ b/doc/man/man3/crypto_x25519.3monocypher
@@ -1,1 +1,81 @@
-crypto_key_exchange.3monocypher
+.Dd December 7, 2017
+.Dt CRYPTO_X25519 3MONOCYPHER
+.Os
+.Sh NAME
+.Nm crypto_x25519
+.Nd X25519 key exchange
+.Sh SYNOPSIS
+.In monocypher.h
+.Ft int
+.Fo crypto_x25519
+.Fa "uint8_t raw_shared_secret[32]"
+.Fa "const uint8_t your_secret_key[32]"
+.Fa "const uint8_t their_public_key[32]"
+.Fc
+.Sh DESCRIPTION
+.Fn crypto_x25519
+computes a shared secret with
+.Fa your_secret_key
+and
+.Fa their_public_key .
+.Sy The shared secret is not cryptographically random .
+Do not use it directly as a session key.
+Hash it with
+.Xr crypto_chacha20_H 3monocypher
+or
+.Xr crypto_blake2b 3monocypher 
+first.
+.Pp
+.Fn crypto_x25519
+is a low-level primitive.
+Users should use
+.Xr crypto_key_exchange 3monocypher
+unless they have a specific reason not to.
+.Sh RETURN VALUES
+Some public keys force the shared key to a known constant.
+This function returns -1 if it detects such a public key, otherwise it
+returns 0.
+This never happens with legitimate public keys, but if the ones you
+process are not known to be trustworthy, check the return value.
+.Sh EXAMPLES
+Generate a pair of shared keys with your secret key and their public
+key.
+(This can help nonce management for full duplex communications.)
+.Bd -literal -offset indent
+const uint8_t their_pk     [32]; /* Their public key          */
+const uint8_t your_sk      [32]; /* Your secret key           */
+uint8_t       shared_secret[32]; /* Shared secret (NOT a key) */
+if (crypto_x25519(shared_secret, your_sk, their_pk) != 0) {
+    /* Their public key is malicious. */
+    /* The exchange must be aborted.  */
+}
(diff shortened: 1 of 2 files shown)


doc/man/man3/crypto_x25519.3monocypher
```diff
@@ -1 +0,0 @@
-crypto_key_exchange.3monocypher
```
mikejsavage: I only just noticed this field, and I haven't been updating it with my other tweaks

is that a problem?

mikejsavage:
bah github won't let me comment on the new file. I guess it's because it used to be a symlink?

attaching screenshot for the convenience of github support so they can see what I mean:

2017-12-07-214423_884x422_scrot

doc/man/man3/crypto_x25519.3monocypher
```diff
@@ -1 +0,0 @@
...
\ No newline at end of file
```
mikejsavage: test

mikejsavage:
[the crypto_x25519 page probably needs a single key example](https://github.com/LoupVaillant/Monocypher/pull/72/files#diff-ec87204b2f937ad6708f288f28304a90R40)

[I'm still not happy with the forward secrecy part.](https://github.com/LoupVaillant/Monocypher/pull/72/files#diff-ec87204b2f937ad6708f288f28304a90R69)

It says you should verify the new shared key with crypto_lock. I suppose that assumes that both ends are controlled by you.

When I first asked about it I had something like HTTPS/game networking in mind where someone else controls the client. If you use crypto_lock and a shared key for authenticating the x25519 public keys then you have to give the secret key to everyone meaning everyone can impersonate the server.

I believe the game networking use case would go something like this:

- hardcode the server's crypto_sign public key into the client
- when forming a new connection, both the client and server generate random x25519 key pairs
- the client sends its x25519 public key to the server raw
- the server crypto_signs its x25519 public key and sends that to the client
- the client verifies it is talking to the correct server
- crypto_lock with the new key as usual

doc/man/man3/crypto_x25519.3monocypher
LoupVaillant: Oh yeah, thanks for the reminder. It's not a problem. I was planing to dedicate a commit to those dates, to mark the day the manual was last reviewed.

CuleX:
Getting same invalid position issue.

* x25519:59: Forgot to `*/`, now it just won't compile as-is.

About the "game networking use case", I believe the idea to use signatures from a pre-shared public key over a key to be fairly obvious, but Loup may disagree.

LoupVaillant:
By "single key example", you mean, generating a public key from a private key? I thought it was obvious enough not to write an example, but we may add that for the sake of completeness.

I see why you're not happy with the forward secrecy section, it is so terse it cannot be understood. I didn't mean what you wrote. Starting from a shared symmetric key would defeat the purpose of a public key setup in the first place: if I controlled both ends and a symmetric key can be shared, I would just use that. (Forward secrecy can be achieved with key erasure: hash the key after the transaction, then use that hash as the new key for the next transaction.)

@CuleX, I agree that signature from pre-shared public keys is a fairly obvious idea that works.

@mikejsavage, your protocol works, and could be mentioned in the manual.  I just have a few qualms:

* We depend on *two* public key algorithms: EdDSA and x25519. I like to minimize my dependencies.
* EdDSA signatures reveal the identity of the server to outside observers. Not a problem for games, but on a peer-to-peer exchange going through onion routing, you don't want an attacker to guess who is talking to whom just by looking at the initial handshake.
* Doing this naively may allow replay attacks. Replay attacks are easy to defend against, but they still require some explaining.

Still, your use case is better than the current forward secrecy part.

Now about what I *actually* had in mind when I talked about authenticating with `crypto_lock`:

* Client and server start with a long term x25519 public key each, known to each other prior to the hand shake.
* Client and server generate a temporary key pair each.
* Client and server perform `key_exchange` with the long term keys. They now have the same long term shared secret.
* Client and server use their long term shared secret to encrypt and authenticate their temporary public keys, which they then send to each other.
* Client and server perform *another* `key_exchange`, this time with the temporary keys.They now  have a temporary shared secret they can use to communicate securely.

The advantage is, the content of the handshake does not reveal the identity of the client or the server to outsiders: everything is encrypted with a shared secret, either long term or temporary. Those identities only get revealed if one of the long term keys gets compromised. There are even ways to avoid this leak by being clever with the first key exchanges.

Problem is, explaining the 3-way handshake required to do this properly would require an entire blog post (I think I will write one). It's too big for the manual.

LoupVaillant:
I know how to solve this: we should replace this section with a reference with a known algorithm. I suggest the Double Ratchet Algorithm.

mikejsavage:
> I thought it was obvious enough not to write an example

most of monocypher is obvious enough to not need examples :)

PR #75, opened 2017-12-29
Misc manual tweaks

mikejsavage:
nothing too exciting

### DIFF
--- a/doc/man/man3/crypto_lock_init.3monocypher
+++ b/doc/man/man3/crypto_lock_init.3monocypher
@@ -235,14 +235,14 @@ XChacha20 derives from Chacha20 the same way XSalsa20 derives from
 Salsa20, and benefits from the same security reduction (proven secure
 as long as Chacha20 itself is secure).
 .Sh SECURITY CONSIDERATIONS
-Using this incremental for decryption means authentication is
-incremental as well.
-This allows two errors: forgetting to call
-.Fn crypto_unlock_final
-altogether, and start processing the message before the authentication
-is complete.
+Messages must be completely decrypted before they can be verified.
+Make sure to call
+.Fn crypto_unlock_final ,
+and make sure to check its return value
+.Em before
+processing the message.
 Messages may be stored before they are verified, but they cannot be
-.Em trusted .
+trusted.
 Processing untrusted messages increases the attack surface of the
 system.
 Doing so securely is hard.
--- a/doc/man/man3/crypto_lock_init.3monocypher
+++ b/doc/man/man3/crypto_lock_init.3monocypher
@@ -235,10 +235,9 @@ XChacha20 derives from Chacha20 the same way XSalsa20 derives from
 Salsa20, and benefits from the same security reduction (proven secure
 as long as Chacha20 itself is secure).
 .Sh SECURITY CONSIDERATIONS
-Messages must be completely decrypted before they can be verified.
-Make sure to call
-.Fn crypto_unlock_final ,
-and make sure to check its return value
+Messages are not verified until the call to
+.Fn crypto_unlock_final .
+Make sure to call it and check the return value
 .Em before
 processing the message.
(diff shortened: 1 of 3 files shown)


doc/man/man3/crypto_sign_init_first_pass.3monocypher
```diff
@@ -167,17 +167,18 @@ These functions implement EdDSA with Curve25519 and Blake2b.
...
 Messages may be stored before they are verified, but they cannot be
 .Em trusted .
 Processing untrusted messages increases the attack surface of the
 system.
 Doing so securely is hard.
 Do not process messages before calling
 .Fn crypto_check_final .
+Additionally, the security considerations documented in
+.Xr crypto_sign 3monocypher ,
+apply here too.
```
LoupVaillant: Actually they do not: hardware fault injection can cause a verification to fail when it should have otherwise succeeded, but the reverse is unlikely. Moreover, this can never lead to key recovery at the verification stage, because we don't process anybody's secrets here.

I think we shouldn't add this sentence.

doc/man/man3/crypto_sign_init_first_pass.3monocypher
```diff
@@ -167,17 +167,18 @@ These functions implement EdDSA with Curve25519 and Blake2b.
 This is the same as Ed25519, with Blake2b instead of SHA-512.
 Ed25519 is described in RFC 7748.
 .Sh SECURITY CONSIDERATIONS
-In addition to those documented on
-.Xr crypto_sign 3monocypher ,
-incremental signature verification encourages the user to start
-processing the message before verification is done.
+Messages are not verified until the call to
+.Xr crypto_check_final .
```
LoupVaillant: Did you mean `.Fn` (function name) instead of `.Xr` (external reference)?

doc/man/man3/crypto_lock_init.3monocypher
```diff
@@ -235,14 +235,14 @@ XChacha20 derives from Chacha20 the same way XSalsa20 derives from
 Salsa20, and benefits from the same security reduction (proven secure
 as long as Chacha20 itself is secure).
 .Sh SECURITY CONSIDERATIONS
-Using this incremental for decryption means authentication is
-incremental as well.
-This allows two errors: forgetting to call
-.Fn crypto_unlock_final
-altogether, and start processing the message before the authentication
-is complete.
+Messages must be completely decrypted before they can be verified.
```
LoupVaillant: Actually, no. Monocypher uses an "encrypt-then-auth" scheme, so you could totally verify before you decrypt. Still, you are right that a reformulation is called for. I vote for something like you wrote for crypto_check:

    Messages are not verified until the call to
    .Fn crypto_unlock_final .
    Make sure to call it and check its return value
    .Em before
    processing the message.

More or less.

doc/man/man3/crypto_sign_init_first_pass.3monocypher
mikejsavage: should it say "when signing messages, the security considerations docmented in crypto_sign also apply"?

doc/man/man3/crypto_sign_init_first_pass.3monocypher
mikejsavage: whoops

doc/man/man3/crypto_sign_init_first_pass.3monocypher
LoupVaillant: Oh, shit, I got confused. I thought for a moment there was a dedicated crypto_check page. Yes, "when signing messages, the security considerations docmented in crypto_sign also apply" is perfect.

LoupVaillant: (To be clear, I retract my earlier comment: the sentence you added was fine.)

mikejsavage:
yeah I'd like to go over it once more

Issue #58, opened 2017-11-12
Docs feedback


mikejsavage:
I'm not done yet, but here are my notes so far. I hope this is readable!

intro.html

- Should explicitly mention that dictionary coders (i.e. every mainstream compression library) leaks info about the plaintext
- Not related to intro.html but perhaps there should be a validation layer that checks things are aligned how they should be/things aren't overlapping when they shouldn't be/etc.

crypto_aead_lock.html
crypto_aead_unlock.html
crypto_lock.html
crypto_unlock.html

- Is it ok to use a counter for the nonce?
- "Make sure message lengths do not leak secret information." - Again be explicit about dictionary coders
- "That length is not authenticated." - I don't fully understand the problem or the solution here. You have to put the length in the authenticated data or it's not secure? Why doesn't monocypher do that for you? It's only 8 bytes.
- In the in place encryption example the cipher_text variable is unused, except for the last comment where it is incorrect.
- Do you need to call crypto_wipe even when crypto_unlock fails? Can it partially decrypt your data?

crypto_argon2i.html

- "It can be used to encrypt private keys or password databases." - "it" is the resulting key?
- "malloc() should yield a suitable chunk of memory." - malloc will (not should) return 8 byte aligned addresses on any normal platform. Might be worth explicitly mentioning that.
- "The version provided by Monocypher has no threading support, so the degree of parallelism is limited to 1. This is considered good enough for most purposes." - I don't know why you mention this
- "This most likely has no practical application but is exposed for the sake of completeness." - TBH I would not expose it in that case
- "Use crypto_verify16(3monocypher), crypto_verify32(3monocypher) or crypto_verify64(3monocypher) to compare password hashes to prevent timing attacks, which are feasible against the weakest passwords." - maybe drop the bit after the comma
- "The hardness of the computation can be chosen thus:" and the two bullet points are oddly phrased. How about:

        To select the nb_blocks and nb_iterations parameters, you should first
        decide how long the computation should take. For user authentication,
        somewhere between half a second (convenient) and several seconds
        (paranoid) is a good starting point, and you should use as much memory
        as you can spare.

        Since parameter selection depends on your hardware, you will need to
        use some trial and error to find the ideal settings. Three iterations
        and 100k blocks (one hundred megabytes of memory) is a good starting
        point. Adjust nb_blocks first. If using all available memory is not
        slow enough, increase nb_iterations.

crypto_blake2b.html
crypto_blake2b_final.html
[...]

PR #76, opened 2018-01-01
More manual tweaks

mikejsavage:
I have tried to avoid minor phrasing changes because that's an infinitely long road that everyone is sick of

I checked every function signature in the docs match monocypher.h, not all of them do

all sizes in args blocks are now, in bytes.

I compiled all the examples as C/C++ with -Wall -Wextra on GCC and /W4 on Windows. W4 gives some junk warnings about type conversions in monocypher.c, but nobody uses W4 (not even microsoft, many standard includes are not W4 clean...). I **didn't** actually run them and test the outputs

@LoupVaillant: to save you from trawling through today's spam on #58 I have added the unresolved issues as comments here

sorry about your inboxes btw.

### DIFF
--- a/doc/man/man3/crypto_argon2i.3monocypher
+++ b/doc/man/man3/crypto_argon2i.3monocypher
@@ -104,12 +104,11 @@ Can be
 if
 .Fa key_size
 is zero.
-The key is generally not needed, but it does have some uses: it is
-supposed to be unknown to the attacker.
-In the context of password derivation, it would stay unknown
-.Em even if an attacker steals the password database .
-This may be possible if that key is stored on a separate server.
-Note: changing the key requires hashing the user's password,
+The key is generally not needed, but it does have some uses.
+In the context of password derivation, it would be stored separately
+from the password database, and would remain secret even if an
+attacker were to steal the database.
+Note that changing the key requires rehashing the user's password,
 which is only possible upon user login.
 .It Fa key_size
 The length of the key, in bytes.
--- a/doc/man/man3/crypto_argon2i.3monocypher
+++ b/doc/man/man3/crypto_argon2i.3monocypher
@@ -50,9 +50,9 @@ This is considered good enough for most purposes.
 The arguments are:
 .Bl -tag -width Ds
 .It Fa hash
-Buffer for the output hash.
+The output hash.
 .It Fa hash_size
-The length of
+Length of
 .Fa hash ,
 in bytes.
 This argument should be set to 16, 32 or 64 for compatibility with the
@@ -86,7 +86,7 @@ It should be wiped with
 .Xr crypto_wipe 3monocypher
 after being hashed.
 .It Fa password_size
-The length of
+Length of
 .Fa password ,
 in bytes.
 .It Fa salt
@@ -94,7 +94,7 @@ A password salt.
 This should be filled with random bytes, generated separately for each
 password to be hashed.
 .It Fa salt_size
-The length of
+Length of
 .Fa salt ,
 in bytes.
 Must be at least 8.
@@ -113,7 +113,9 @@ attacker were to steal the database.
 Note that changing the key requires rehashing the user's password,
 which is only possible upon user login.
 .It Fa key_size
-The length of the key, in bytes.
+Length of
+.Fa key ,
+in bytes.
 Must be zero if there is no key.
 .It Fa ad
 Additional data.
@@ -128,7 +130,7 @@ if
 .Fa ad_size
 is zero.
 .It Fa ad_size
-Ahe length of
+Length of
 .Fa ad ,
 in bytes.
 .El
@@ -144,8 +146,12 @@ or
 .Xr crypto_verify64 3monocypher
 to compare password hashes to prevent timing attacks.
 .Pp
-To select the nb_blocks and nb_iterations parameters, it should first be
-decided how long the computation should take.
+To select the
+.Fa nb_blocks
+and
+.Fa nb_iterations
+parameters, it should first be decided how long the computation should
+take.
 For user authentication, we recommend somewhere between half a second
 (convenient) and several seconds (paranoid).
 The computation should use as much memory as can be spared.
--- a/doc/man/man3/crypto_argon2i.3monocypher
+++ b/doc/man/man3/crypto_argon2i.3monocypher
@@ -52,9 +52,7 @@ The arguments are:
 .It Fa hash
 Buffer for the output hash.
 .It Fa hash_size
-The length of the output
-.Fa hash
-in bytes.
+The length of the output hash, in bytes.
 This argument should be set to 16, 32 or 64 for compatibility with the
 .Fn crypto_verify*
 constant time comparison functions.
@@ -86,20 +84,21 @@ It should be wiped with
 .Xr crypto_wipe 3monocypher
 after being hashed.
 .It Fa password_size
-the length of
-.Fa password .
+The length of
+.Fa password ,
+in bytes.
 .It Fa salt
-a password salt.
+A password salt.
 This should be filled with random bytes, generated separately for each
 password to be hashed.
 .It Fa salt_size
-the length of
+The length of
 .Fa salt ,
 in bytes.
 Must be at least 8.
 16 is recommended.
 .It Fa key
-a key to use in the hash.
+A key to use in the hash.
 Can be
 .Dv NULL
 if
@@ -113,10 +112,10 @@ This may be possible if that key is stored on a separate server.
 Note: changing the key requires hashing the user's password,
 which is only possible upon user login.
 .It Fa key_size
-the length of the key, in bytes.
+The length of the key, in bytes.
 Must be zero if there is no key.
 .It Fa ad
-additional data.
+Additional data.
 This is additional data that goes into the hash, similar to the
 authenticated encryption with authenticated data (AEAD) construction in
 .Xr crypto_aead_lock 3monocypher .
@@ -128,8 +127,9 @@ if
 .Fa ad_size
 is zero.
 .It Fa ad_size
-the length of
-.Fa ad .
+Ahe length of
+.Fa ad ,
+in bytes.
 .El
 .Pp
 The output hash must not overlap with the work area, or it will be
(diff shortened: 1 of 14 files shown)


mikejsavage:
args blocks are somewhat inconsistent between "The length of the output hash in bytes." and "The length of the output .Fa hash in bytes."

I think it should either by the first or "The length of .Fa hash in bytes." but i don't care that much

any opinions?

I changed "The length of the output .Fa hash in bytes" to not have the .Fa and leaving the rest as is. CuleX voted for using .Fa more. If you agree I can do another pass over the args blocks

mikejsavage:
last arg to void crypto_chacha20_stream is called size in the header, stream_size in the manual

mikejsavage:
crypto_lock_auth args are called ad/ad_size in the manual, message/message_size in the header

I believe the manual is correct here as crypto_lock_auth deals with additional data

mikejsavage:
crypto_wipe's size arg is called secret_size in the manual, size in the header

also I guess "It probably wasn't the only copy..." in the header is a joke but it's a bit confusing :d

mikejsavage:
crypto_aead_lock/argon2i examples are like

uint8_t       *text;      /* Message to decrypt        */
size_t         text_size; /* Message size (NOT secret) */

when other pages have examples like uint8_t message[500];

should we fix the examples to only use one or the other?

doc/man/man3/crypto_sign.3monocypher
```diff
@@ -78,8 +78,10 @@ message.
 It does not have to in most threat models, because nothing is secret:
 everyone knows the public key, and the signature and message are
 rarely secret.
-To ascertain the origin of a secret message, use
+If the message needs to be secret, use
```
mikejsavage: I don't think the original text was super clear here

doc/man/man3/crypto_lock.3monocypher
```diff
@@ -82,9 +82,9 @@ See
 about random number generation (use your operating system's random
 number generator).
 .It Fa mac
-a 16-byte
-.Em message authentication code (MAC) ,
-that can only be produced by someone who knows the session key.
+A 16-byte
+.Em message authentication code
+(MAC), that can only be produced by someone who knows the session key.
```
mikejsavage: "message authentication code" in italics, (MAC) not. I think this is the normal way of doing it

doc/man/man3/crypto_key_exchange.3monocypher
```diff
@@ -43,13 +43,11 @@ Do not use the same private key for both key exchanges and signatures.
...
 The public key of the other party.
-.It Fa public_key
-The public key, generated from the
+.It Fa your_public_key
+Your public key, generated from
 .Fa secret_key
 with
 .Fn crypto_x25519_public_key .
-.It Fa secret_key
-A 32-byte random number, known only to you.
```
mikejsavage: I was surprised to see secret key appear twice!

If you agree maybe monocypher.h could be changed to match the docs too

doc/man/man3/crypto_chacha20_encrypt.3monocypher
```diff
@@ -273,9 +268,9 @@ crypto_chacha20_encrypt(&ctx,
 crypto_chacha20_set_ctr(&ctx, 0);
 crypto_chacha20_encrypt(&ctx, cipher_text, plain_text, 3 * 64);
 /* Wipe secrets if they are no longer needed */
-crypto_wipe(key,        sizeof(key));
+crypto_wipe(key,        32);
 crypto_wipe(&ctx,       sizeof(ctx));
-crypto_wipe(plain_text, sizeof(plain_text));
```
mikejsavage: CuleX: "I'm certain that we've agreed some time ago to avoid sizeof for the examples"

doc/man/man3/intro.3monocypher
```diff
@@ -195,9 +194,9 @@ are not compromised even if the long term keys are.
 This is done by generating temporary keys, then encrypting messages
 with them.
 .Pp
-This can be generalised: secrets that went through a computer should
-not be compromised when this computer is stolen or infected at a later
-point.
+.Ss Temporary secrets
```
mikejsavage: not sure about this change

mikejsavage:
intro says "Recent versions of Linux (glibc >= 2.25, Linux >= 3.17), provide .Fn getrandom"

getentropy (glibc 2.25 also) is a tiny bit easier to use (no EINTR or EAGAIN). should we mention that?

mikejsavage:
crypto_aead_lock/argon2i examples are like

```
uint8_t       *text;      /* Message to decrypt        */
size_t         text_size; /* Message size (NOT secret) */
```

when other pages have examples like `uint8_t message[500];`

should we fix the examples to only use one or the other?

CuleX:
Avoid mentioning `getentropy`; the Linux man page for it says "It is also present on OpenBSD" and that it's implemented using getrandom (presumably with the wrong arg), OpenBSD says: "[getentropy() is not intended for regular code; please use the arc4random(3) family of functions instead.](https://man.openbsd.org/getentropy)"

If getentropy on Linux is implemented using getrandom() flag `GRND_RANDOM`, it goes against our own advice (call with no flags set) because (1) it can block, (2) it yields raw entropy rather than random bytes generated using that entropy. Plus OpenBSD is very stern about not wanting to use it; people may wrongly think they can use getentropy as a (more) portable RNG.

mikejsavage:
ugh comments here don't have threads either, only reviews. now this is an even bigger mess

should `crypto_check_public_key` be documented?

LoupVaillant:
Man, the big red circle we have when I request changes… A bit aggressive, don't you think?

> crypto_lock_auth args are called ad/ad_size in the manual, message/message_size in the header

The reason for the discrepancy is because in the manual, we only recommend using `crypto_lock_auth()` with additional data. In reality however, it is used to authenticate *everything*, including the message. See the source code for `crypto_lock_update()` and `crypto_unlock_update()`. Thus, I vote for keeping the discrepancy.

> last arg to void crypto_chacha20_stream is called size in the header, stream_size in the manual  
> crypto_wipe's size arg is called secret_size in the manual, size in the header

This is because in the header, we can infer which size we are referring to from context. I tend to prefer concision when it doesn't hurt, so I'd rather keep the discrepancy.

> I guess "It probably wasn't the only copy..." in the header is a joke but it's a bit confusing :d

While it was meant tongue in cheek, I did want to point to a real problem: stuff tends to be copied everywhere, and secrets are no exception. The user should make a reasonable attempt at erasing *every* such copy, not just the one that lays before their eyes. I'm not against a more serious "make sure you erase every copy" or something if you think it is warranted. I like my jokes, but clarity is more important.

> getentropy

I trust @CuleX on this one. Don't mention it.

> `crypto_lock_encrypt()`

The header file doesn't explicitly says anything is a low-level primitive, so I haven't thought of putting any warning for this one. We should probably add warnings for low-level primitives in the header, if only to deter users who don't read the manual.

doc/man/man3/crypto_chacha20_encrypt.3monocypher
```diff
@@ -71,17 +71,19 @@ The message to encrypt.
...
 .Fa plain_text and
-.Fa cipher_text .
+.Fa cipher_text ,
+in bytes.
 .It Fa stream
-the raw Chacha20 stream.
+The raw Chacha20 stream.
 .It Fa stream_size
-The size of the stream.
+The size of the stream, in bytes.
```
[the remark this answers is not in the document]
mikejsavage: size, or length :d

also the or no the

doc/man/man3/crypto_lock.3monocypher
```diff
@@ -203,12 +204,12 @@ crypto_wipe(key, 32);
...
 .Bd -literal -offset indent
-const uint8_t  key  [32];    /* Same as the above         */
-const uint8_t  nonce[24];    /* Same as the above         */
-const uint8_t  mac  [16];    /* Received from the network */
-const uint8_t *cipher_text;  /* Encrypted message         */
-size_t         text_size;    /* Message size (NOT secret) */
-uint8_t       *plain_text;   /* Secret message            */
+const uint8_t  key  [32];   /* Same as the above         */
+const uint8_t  nonce[24];   /* Same as the above         */
+const uint8_t  mac  [16];   /* Received from the network */
```
mikejsavage: is "Received from the network" right here?

mikejsavage:
> I'm not against a more serious "make sure you erase every copy" or something if you think it is warranted. I like my jokes, but clarity is more important.

I'd say it's worth mentioning, you can even keep the joke like "It probably wasn't the only one... make sure you wipe every copy!"

> if only to deter users who don't read the manual.

it's safe to say that i have read the manual at this point, and i can still see myself forgetting, looking in the header, and picking the wrong one :d

mikejsavage:
github has nicely buried my comments, see https://github.com/LoupVaillant/Monocypher/pull/76#discussion_r159523327 and https://github.com/LoupVaillant/Monocypher/pull/76#discussion_r159522600

doc/man/man3/crypto_argon2i.3monocypher
```diff
@@ -52,7 +52,9 @@ The arguments are:
 .It Fa hash
 Buffer for the output hash.
 .It Fa hash_size
-The length of the output hash, in bytes.
+The length of
+.Fa hash ,
+in bytes.
```
LoupVaillant: Yes, that's perfect.

doc/man/man3/crypto_chacha20_encrypt.3monocypher
```diff
@@ -83,7 +83,9 @@ in bytes.
 .It Fa stream
 The raw Chacha20 stream.
 .It Fa stream_size
-The size of the stream, in bytes.
+The size of
+.Fa stream ,
+in bytes.
```
LoupVaillant: I personally vote for "length" and no "the". So,

    Length of
    .Fa stream ,
    in bytes.

Though I care more about consistency than this particular phrasing.

doc/man/man3/crypto_chacha20_encrypt.3monocypher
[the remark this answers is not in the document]
LoupVaillant: Oh, crap.

doc/man/man3/crypto_lock.3monocypher
LoupVaillant: Strictly speaking, any unsecure channel would do, it doesn't have to be the network specifically. The network is just a typical example. I don't know a concise way to say it, though. "Received from the sender", maybe? "Sent by the sender"? I think "Received from the network" is good enough, but we can change it if you prefer something else.

LoupVaillant:
> "It probably wasn't the only one... make sure you wipe every copy!"

Good idea, I'll do that.

> it's safe to say that i have read the manual at this point, and i can still see myself forgetting, looking in the header, and picking the wrong one :d

Indeed, and the user will end up with an unauthenticated messages, possibly without noticing it in the tests (`crypto_lock_encrypt()` is symmetric). Major oops. I'll add a warning about the function not being low-level, and not intended for end users.

doc/man/man3/crypto_lock.3monocypher
mikejsavage: I don't mind it but you guys were talking about it on one of the other PRs

LoupVaillant: Bah, I'd rather leave it alone it looks good enough.

PR #92, opened 2018-03-21
Fix copy-paste mistake in intro docs

SgtCoDFish:
This is a very small copy paste error I noticed in https://monocypher.org/manual/#Timing_attacks


LoupVaillant:
Good catch, thank you!

PR #95, opened 2018-04-18
fix include when using ED25519_SHA512

vbmithr:


LoupVaillant:
Just to confirm, is this because you get an incomplete type error when using the incremental signing interface?

vbmithr:
No, I just can't compile without it.
```LANG=C gcc -Ioptional -DED25519_SHA512 -c optional/sha512.c monocypher.c``` does not work without my patch. Is there something wrong in the way I compile it?

vbmithr:
np :) Just try compiling the code with my command-line above and you will probably see an error, which is likely to be the same as what I have (missing definitions, blah blah… the error that happens when you forget to insert a `.h` file).

LoupVaillant:
Oh, I was confused by the leading `LANG=C`. Anyway, I see the problem now. Accepting the PR right away, thanks.

Issue #102, opened 2018-07-01
Return code or no return code for key exchange?


LoupVaillant:
In its early stage, Monocypher's key exchange primitives didn't have a return value. It has since [been introduced](https://github.com/LoupVaillant/Monocypher/commit/b7826a99ef11b6a65db251b872bf413dcb9c87f5#diff-b7175b069a24cabf8d434251283462fcL105), on the grounds that contributory behaviour matters, and Libsodium uses it (TweetNaCl doesn't).

I've always had a problem with this test. When both keys come from legitimate sources (as they do when two honest parties that know of each other's public keys beforehand), then there is no *need* for such a test. And I don't like to complicate the interface by adding a significant return value to a primitive that doesn't need one. This is also contrary to Monocypher's general philosophy of validating nothing (it's the user's responsibility to provide correct inputs). Besides, DJB himself says [we shouldn't validate](http://cr.yp.to/ecdh.html#validate). Still, this seemed small enough that I could just add it.

[This argument](https://moderncrypto.org/mail-archive/curves/2017/000896.html) by Trevor Perrin changed my mind:

- The check does have a cost: it's a conditional branch, and crypto must be cautious with them.
- When needed, the check is insufficient. If for instance the public key is a known multiple of the base point (such as 37, as suggested in the original paper), then anyone can reconstruct the shared secret. There is no reasonable way to test for this.
- To simplify the security analysis, we should only provide points that are on the curve. This means not on its twist. Validating *that* is way more expensive than just checking for all zero at the end.
- Most of all, such validation endorses protocols that rely on properties *outside the scope of Diffie Hellman key exchange*.

That last one is probably the most harmful. While curve25519 may have interesting properties besides ensuring DH works as intended, relying on those other properties makes a brittle protocol that may break if we change the underlying primitive. Also, those other security properties have probably received less scrutiny than the main one.

So I'm thinking about removing the check, incurring a minor breaking change in the API, thus bumping the version to 2.1.0. Any thought?

dsprenkels:
Just looking at your library for the first time, so feel free to ignore this comment. My two cents:

Diffie-Hellman must *always* be authenticated, and thus the inputs *are always* valid. Even if they are not, the exception should not occur in the Diffie-Hellman exchange, but in the authentication check. Adding a check in the scalar multiplication would add a (timing) oracle.

LoupVaillant:
[...]

PR #105, opened 2018-07-17
Deprecate return values

fscoto:
This addresses #102.


LoupVaillant:
Nice, that's exactly what we needed, thanks.

LoupVaillant:
Oh, come to think of it, I believe the same should be done for `crypto_x25519()`.

PR #106, opened 2018-07-18
Deprecate return values in crypto_x25519

fscoto:
This addresses #102.

This also adds a note that the public key function returns nothing, as is noted on crypto_key_exchange(3monocypher).


LoupVaillant:
Thanks!

Issue #89, opened 2018-02-10
2.0.0 release candidate


LoupVaillant:
I have pushed [the new AEAD interface](https://github.com/LoupVaillant/Monocypher/commit/09de649d537e963dbb6b7a6f1bc6885f16df5078). Hopefully it will be really stable this time. (I believe it will be because it is now compatible with Libsodium's `crypto_aead_xchacha20poly1305_ietf_encrypt()`, to the point where I was able to make test vectors.)

Since this change breaks everything, I took the opportunity to tweak the incremental interface. The subsequent compile time errors may help users figure out why their data isn't decrypting itself any more. Or maybe it will just piss them off. I'm not sure.

I'm thinking of renaming  `crypto_aead_lock()` into `crypto_lock_aead()` (and `crypto_aead_unlock` into `crypto_unlock_aead`). This would have 2 effects:

* Naming wise, this would make everything start by `crypto_lock` or `crypto_unlock`. This may or may not matter. I just like being consistent, and this was a tiny inconsistency.
* Versioning wise, this change will trigger compile time errors for virtually every user. Is signalling the incompatibility this way actually a good idea, or will it just piss them off, and I'd better not force them to change their source code? (I know I already did this for the incremental interface, but I expect virtually no one had the time to actually use it yet. Doing the same to the monolithic interface will affect virtually everyone.)

@CuleX, @mikejsavage, @secworks, @konovod, what do you think?

secworks:
I think following/using the construct from RFC7539 is a really good move.

konovod:
Telling in readme about incompatibility is IMHO enough, no need to change names. 
On the other hand, even big libraries like to change names, I don't think that can piss off anyone. So if renaming is useful for consistency - do it.

LoupVaillant:
@secworks, note that I did make a small departure from RFC7539, namely using XChacha20 instead of the IETF version of Chacha20 (with a 96-bit nonce and 32-bit counter). This effectively means processing an additional Chacha block for each message. But I judged this overhead was worth the extra convenience and security that comes with being able to use random nonces. (Counters are stateful. Random stuff is effectively stateless, and thus leaves less room for error.)

@konovod, thanks. I'll probably change the names before release, unless someone else points out a good reason not to.

CuleX:
[...]

PR #122, opened 2019-04-03
Add HISTORY sections

fscoto:
These changes make it easier for people who read a newer manual than the one they have. This could happen if a long-term stability Linux distribution ends up packaging Monocypher and backporting security fixes, but developers are reading the manual on the web.

The wording of each first sentence matches OpenBSD's for section 3 HISTORY sections.

### DIFF
--- a/doc/man/man3/crypto_lock.3monocypher
+++ b/doc/man/man3/crypto_lock.3monocypher
@@ -249,3 +249,13 @@ These functions implement RFC 8439, with XChacha20 instead of Chacha20.
 XChacha20 derives from Chacha20 the same way XSalsa20 derives from
 Salsa20, and benefits from the same security reduction (proven secure
 as long as Chacha20 itself is secure).
+.Sh HISTORY
+The
+.Fn crypto_lock
+and
+.Fn crypto_unlock
+functions first appeared in Monocypher 0.1.
+.Fn crypto_lock_aead
+and
+.Fn crypto_unlock_aead
+were introduced in Monocypher 1.1.0.
--- a/doc/man/man3/crypto_lock.3monocypher
+++ b/doc/man/man3/crypto_lock.3monocypher
@@ -259,3 +259,12 @@ functions first appeared in Monocypher 0.1.
 and
 .Fn crypto_unlock_aead
 were introduced in Monocypher 1.1.0.
+In Monocypher 2.0.0, the underlying algorithms for these functions were
+changed from a custom XChacha20/Poly1305 construction to an
+implementation of RFC 7539 (now RFC 8439) with XChacha20 instead of
+Chacha20.
+The
+.Fn crypto_lock_encrypt
+and
+.Fn crypto_lock_auth
+functions were removed in Monocypher 2.0.0.
--- a/doc/man/man3/crypto_lock_init.3monocypher
+++ b/doc/man/man3/crypto_lock_init.3monocypher
@@ -273,6 +273,20 @@ These functions implement RFC 8439, with XChacha20 instead of Chacha20.
 XChacha20 derives from Chacha20 the same way XSalsa20 derives from
 Salsa20, and benefits from the same security reduction (proven secure
 as long as Chacha20 itself is secure).
+.Sh HISTORY
+The
+.Fn crypto_lock_init ,
+.Fn crypto_lock_auth_ad ,
+.Fn crypto_lock_auth_message ,
+.Fn crypto_lock_update ,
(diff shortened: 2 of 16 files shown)


doc/man/man3/crypto_lock.3monocypher
```diff
@@ -249,3 +249,13 @@ These functions implement RFC 8439, with XChacha20 instead of Chacha20.
...
+.Sh HISTORY
+The
+.Fn crypto_lock
+and
+.Fn crypto_unlock
+functions first appeared in Monocypher 0.1.
+.Fn crypto_lock_aead
+and
+.Fn crypto_unlock_aead
+were introduced in Monocypher 1.1.0.
```
LoupVaillant: Perhaps mention that a breaking change was introduced in version 2.0.0?

doc/man/man3/crypto_lock_init.3monocypher
```diff
@@ -273,6 +273,20 @@ These functions implement RFC 8439, with XChacha20 instead of Chacha20.
...
+.Fn crypto_lock_auth_message ,
+.Fn crypto_lock_update ,
+.Fn crypto_lock_final ,
+.Fn crypto_unlock_init ,
+.Fn crypto_unlock_auth_ad ,
+.Fn crypto_unlock_auth_message ,
+.Fn crypto_unlock_update ,
+and
+.Fn crypto_unlock_final
+functions first appeared in Monocypher 1.1.0.
```
LoupVaillant: Ditto here (breaking change in version 2.0.0

doc/man/man3/crypto_sign.3monocypher
```diff
@@ -138,6 +138,13 @@ if (crypto_check(signature, pk, message, 500)) {
 These functions implement PureEdDSA with Curve25519 and Blake2b, as
 described in RFC 8032.
 This is the same as Ed25519, with Blake2b instead of SHA-512.
+.Sh HISTORY
+The
+.Fn crypto_sign ,
+.Fn crypto_check ,
+and
+.Fn crypto_sign_public_key
+functions appeared in Monocypher 0.2.
```
LoupVaillant: Perhaps we should mention the critical vulnerability in `crypto_check()`? That vulnerability was introduced in version 0.3 I believe (when I started to perform the scalar multiplication in Montgomery space), and was corrected on versions 2.0.4 and 1.1.1.

doc/man/man3/crypto_sign_init_first_pass.3monocypher
```diff
@@ -164,6 +164,17 @@ if (crypto_check_final(&ctx)) {
...
+The
+.Fn crypto_sign_init_first_pass ,
+.Fn crypto_sign_update ,
+.Fn crypto_sign_final ,
+.Fn crypto_sign_init_second_pass ,
+.Fn crypto_check_init ,
+.Fn crypto_check_update ,
+and
+.Fn crypto_check_final
+functions first appeared in Monocypher 1.1.0.
```
LoupVaillant: (Same thing about the vulnerability)

fscoto:
> One last nitpick: you mention when each function first appeared, even when that version is below 1.0.0. I thought we could maybe simplify things a bit and consider 1.0.0 like it was the beginning of time? This would leave us 2 obvious alternatives:
> 
>     * Remove the history section for functions that appeared before (or at) version 1.0.0, and never changed since.
> 
>     * Or Just mention that those function where "present from the beginning", or "appeared before version 1.0.0".
> 
> 
> I think it would be slightly simpler, but I don't know the best practices. (Also, I don't have a strong opinion on the matter, just a slight instinctive preference).

The reason I went back to before 1.0 is because versions before 1.0 are listed on the downloads page. As long as those old versions are there in historical interest. I'm honestly quite undecided now that you brought it up, though I do have a slight preference for being honest and say “appeared before version 1..0.0” rather than removing history.

doc/man/man3/crypto_sign.3monocypher
fscoto: Looking at some of the unpatched bindings that have been floating around, that might not actually be a bad idea.

doc/man/man3/crypto_lock.3monocypher
fscoto: I actually forgot about commit 09de649d and #89. Oops. Just shows how long ago we should've started with these sections.

LoupVaillant:
OK, let's keep an accurate history. I wasn't sure to begin with, so let's preserve the information for now.

PR #128, opened 2019-10-18
Clean up kex documentation removal

fscoto:
Related to git commit 6163d8195a3acf2e143d20843a602fd5fb7671d5.

It doesn't seem reasonable to ship a manual referencing functions that don't exist.


LoupVaillant:
Oops, forgot about that. Thanks.

PR #129, opened 2019-10-21
Fix release.sh to not clobber __git__ substitution

fscoto:
While already there, use find -exec (available since 7th Edition UNIX) to avoid issues with spaces in filenames or other weirdness related to whitespace in filenames later down the line.


fscoto:
Independently fixed in 5c283223fb092bf41ec081d0f259cf6c0bb0df3e.

PR #130, opened 2019-10-21
Change release.sh to use find -exec

fscoto:
This avoids some potential weirdness with whitespace in find and for.

Probably not going to be an actual concern, but it's just a nice general improvement.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -23,10 +23,7 @@ rm -rf $FOLDER
 rsync -ad --exclude-from=tarball_ignore ./ $FOLDER
 
 # Replace version markers by the actual version number (from tags)
-for file in `find monocypher-$VERSION -type f`
-do
-    sed -i "s/__git__/$VERSION/g" $file
-done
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -23,10 +23,7 @@ rm -rf $FOLDER
 rsync -ad --exclude-from=tarball_ignore ./ $FOLDER
 
 # Replace version markers by the actual version number (from tags)
-for file in `find monocypher-$VERSION -type f`
-do
-    sed -i "s/__git__/$VERSION/g" $file
-done
+find $FOLDER -type f -exec sed -i "s/__git__/$VERSION/g" \{\} \;
```
LoupVaillant: When I type `find -exec` command, I usually don't escape the `{}`. Any particular this escape might be needed? (Not worth fixing if it isn't, just curious.)

fscoto: Some shells might be picky about it since `{` and `}` may have meaning. It's just an abundance of caution and probably not strictly necessary.

LoupVaillant: Okay, I'll keep that in mind in the future.

PR #132, opened 2019-11-08
Added PlatformIO file to make it easier to depend on monocrypt

DavyLandman:
For a nice IoT project I'm working on, I'm going with Monocypher, is has just the right stuff in there.

A lot of people are now using PlatformIO to help with dependency management and compilation for all kinds of boards. The only other cryptography library on there is ArduinoCrypto, it is limited to arduino, quite big, slowish, and doesn't contain the opinionated api that Monocypher and friends adhere to. I think it would be a great portable addition, and PlatformIO could give some users for Monocypher.

The PlatformIO scanner takes care of it all (after initial submission), and only when you bump the version in the version field, will it look for the tag in the repo to find which sourcecode corresponds to that specific version.

I've tested the below config, and used the `BLAKE2_NO_UNROLLING` since platform IO is mostly on limited platforms. Although that might be to restricting of a choice.

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,0 +1,15 @@
+{
+    "name": "Monocypher",
+    "description": "An easy to use, easy to deploy crypto library. It provides functions for authenticated encryption, hashing, password key derivation, key exchange, and public key signatures.",
+    "keywords":  "encryption,signing,key-exchange,password-key-derivation",
+    "version": "2.0.8",
+    "repository": "https://github.com/LoupVaillant/Monocypher",
+    "build": {
+        "srcDir": "src/",
+        "includeDir": "src/",
+        "flags": [
+            "-DBLAKE2_NO_UNROLLING"
+        ],
+        "libLDFMode": "off"
+    }
+}


DavyLandman:
Maybe hold off a bit on accepting, it might be possible to do this in a companion repo, that also contains pio specific examples and documentation. I will do some reading.

Except if you let me know that you have something against putting monocypher on pio.

LoupVaillant:
Thanks for the contribution, but I think you are trying to make a *package* of Monocypher, right?

Monocypher so far limited itself to just the Makefile necessary to build and install it on typical UNIX platforms. Other forms of distributions such as alternate build systems, packaging for specific platforms or distributions, and language bindings, so far happened outside. I usually link the ones I'm aware of from the [download page](https://monocypher.org/download/).

Is there a central repository of PlatformIO libraries? If you can get to register Monocypher there, I would happily link to that under the "alternate distributions" set of links.

library.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "Monocypher",
+    "description": "An easy to use, easy to deploy crypto library. It provides functions for authenticated encryption, hashing, password key derivation, key exchange, and public key signatures.",
+    "keywords":  "encryption,signing,key-exchange,password-key-derivation",
+    "version": "2.0.8",
```
LoupVaillant: Monocypher's latest version is "2.0.6". Typo?

Also, the tarball generator scripts automatically scans for "__git__", and replaces that with the actual version. If I were to accept that pull request, we'll need to write

        "version": "__git__",

DavyLandman: ouch, indeed, 2.0.6 :)

platformio works in a way that this file is retreived from the VCS, and then it points to either the local directory, or a specific GH repo, and then retreives the release taged with `2.0.6` using the GH api. So this file is like a fixed pointer to the last release, I agree, a bit annoying, but if you remove the version field, it publishes all commits, not just the tags you pick.

DavyLandman:
okay, I see your point of not cluttering this repo with repository specific configurations.

just to summarize, [PlatformIO](https://platformio.org/) is (amongst other things) an open source [repository](https://platformio.org/lib) of libraries suited for IoT [boards](https://platformio.org/boards), [frameworks](https://platformio.org/frameworks), and [platforms](https://platformio.org/platforms). It takes care of a lot of messy cross compilation and debugging stuff, as finally making an cross OS c packaging system.

I think I'll try to make the addition to PlatformIO as a seperate repository, to keep this one as clean as possible.

LoupVaillant:
Cool, let me know when it's done, I'll add it to the "alternate distributions" section of the download page. More repositories & packages are always welcome. :smile:

DavyLandman:
Hi @loupvalliant, it's live here: https://platformio.org/lib/show/6830/Monocypher

The meta data lives here: https://github.com/DavyLandman/monocypher-platformio/blob/master/library.json

What would be a good process for version bumps? Is there a mailing list, or do you prefer to submit a pull request? I'm option for suggestions.

LoupVaillant:
OK thanks for the package. There's a couple problems though:

- The `.tar.gz` is a zip bomb: by default, it extracts everything in the current directory instead of creating a dedicated subdirectory that has the same name as the archive (minus the `.tar.gz` suffix). Not a big deal, but quite inconvenient.

- The [list of headers](https://platformio.org/lib/show/6830/Monocypher/headers) is wrong. Monocypher has _one_ exported header `src/monocypher.h`. The `src/optional/sha512.h` is only used to compile Monocypher in Ed25519 compliant mode, and is not really part of the public API. The other headers are only used for the test suite or the speed benchmark. They are not part of the library itself.

- The archive does not contain the right files to begin with. Instead of taking the released files from the tarball, it takes them from the repository. The repository is not intended for end users, but for contributors. Picking files from git instead of using the latest tarball causes several problems:

   - The version in the source file is wrong. It says `__git__` instead of `2.0.6`. Version numbers are important, because they let us check at a glance which version we are actually using. When we see `__git__`, we don't know. When we see `2.0.6`, we know exactly what we're dealing with.

   - The `tests/vectors.h` header is missing. Users won't be able to run the test suite. Most users are liable to think this is unacceptable (and I'd agree with them).

   - The html documentation is missing. Your users only have the man pages.

   - Many useless files are included: `.gitignore`, `tarball_ignore`, the `test/gen/` folder (used to generate `vectors.h`), the `tests/externals/ed25519-donna` folder…

I guess much of this is automated, but the result simply does not work. If PlatformIO needs a git repository (instead of a URL to an official archive), we have to give it to them. Create an empty repository, dump the contents of an official archive, and have the manifest file point to *that*. You can even get around the zip bomb problem by dumping all files into a `monocypher-2.0.6` subfolder (unless of course zip bombs are a well known convention of PlatformIO). I guess the list of headers will still be wrong, but right now we have much bigger problems.

If you could fix those problems that would be wonderful. IoT is an important niche for Monocypher, I'd very much like to list that package officially.

DavyLandman:
>     * The `.tar.gz` is a zip bomb: by default, it extracts everything in the current directory instead of creating a dedicated subdirectory that has the same name as the archive (minus the `.tar.gz` suffix). Not a big deal, but quite inconvenient.

With platformio, you are not suppose to download the tar itself, the platform takes care of unpacking and putting it in the right point.

>     * The [list of headers](https://platformio.org/lib/show/6830/Monocypher/headers) is wrong. Monocypher has _one_ exported header `src/monocypher.h`. The `src/optional/sha512.h` is only used to compile Monocypher in Ed25519 compliant mode, and is not really part of the public API. The other headers are only used for the test suite or the speed benchmark. They are not part of the library itself.

Agreed, I have to work on the meta data, so that it only targets monocypher and not the rest. It works, but it's to big. I especially have to figure out how to do a build-flag specific inclusion of files. I have to ask the people of platformio how to do that.

>     * The archive does not contain the right files to begin with. Instead of taking the released files from the tarball, it takes them from the repository. The repository is not intended for end users, but for contributors. Picking files from git instead of using the latest tarball causes several problems:

Agreed, I'm going to see if I can get it away from a pointer to github, and just use the releases from the site. I wrongly thought that the git extracted would use the releases API, but it just retrieves the commit at the specific tag.


>       * The `tests/vectors.h` header is missing. Users won't be able to run the test suite. Most users are liable to think this is unacceptable (and I'd agree with them).

I Understand your point, I have to check how this fits the model of platformio, since we don't want to compile the test code into the library that is linked into the main IoT code, as we want to reduce binary size.

>       * The html documentation is missing. Your users only have the man pages.

I do not think most users of platformio look at the files behind the library (more than the headers the IDE providers). I did link the website in the meta-data.

> If you could fix those problems that would be wonderful. IoT is an important niche for Monocypher, I'd very much like to list that package officially.

I'll see if I can allocate some time to improve all this.

LoupVaillant:
> With platformio, you are not suppose to download the tar itself, the platform takes care of unpacking and putting it in the right point.

Okay, so the other points would be less of a problem. Especially the absence of the `vectors.h` header. Still, PlatformIO *does* allow users to get a tarball, so it would be nice if it matches the official one. If PlatformIO can use a released archive instead of a git repository as a template, that would indeed be ideal.

That said, departures from the official tarballs (zip bomb, adding the manifest file…) are perfectly okay. I just think they should be deliberate. For instance, if you omit the html documentation, maybe you want to omit *all* documentation? If you omit `vectors.h`, maybe you want to omit *all* the test suite? That kind of deliberate.

LoupVaillant:
Closing this PR, since external packages should stay external (I may however accept light upstream patches to make the work of package maintainers easier).

I'm still looking forward to a nice PlatformIO package, though. :smile:

DavyLandman:
Completely agree, hope to get some time this weekend to make a nicer version of it.

PR #133, opened 2019-11-19
Remove unnecessary dependency on 2's complement

michaelforney:
Just a minor thing I noticed. 2's complement will be the only signed integer representation in C2X (http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2412.pdf), but since Monocypher targets C99, I think it would be good to avoid relying on it.


LoupVaillant:
Good point, thanks. This may trigger some Clang warning, though, I'll check that out.

LoupVaillant:
Err, turned out the error was the comment itself. 2's complement is guaranteed when the integer has a fixed width, so there was no need to say that we "rely on it". It's fully portable C99.

I reverted the `i32 -> u32` change, but I preserved the deletion of the comment.

PR #134, opened 2019-11-19
Fix makefile rule when test/vectors.h is missing

michaelforney:
`return` only works in functions, so when I run `make check` with no
tests/vectors.h, I get

	return 1
	make: return: Command not found
	make: *** [makefile:189: tests/vectors.h] Error 127

While I guess this does the job, the right thing to do here is `exit 1`.

### DIFF
--- a/makefile
+++ b/makefile
@@ -186,7 +186,7 @@ tests/vectors.h:
 	@echo " need libsodium."
 	@echo "======================================================"
 	@echo ""
-	return 1
+	exit 1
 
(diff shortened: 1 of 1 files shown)


LoupVaillant:
Good catch, thanks. It was working in my machine, but I guess `exit` is much more portable.

LoupVaillant:
(Hmm, perhaps it was *not* working on my machine, and I just didn't notice…)

PR #135, opened 2019-11-21
Add SONAME to shared library

fscoto:
This is my oversight and probably should've been done ages ago. The makefile never *actually* added the SONAME information into the shared library.

`$ objdump -p lib/libmonocypher.so` actually contains a `SONAME` after recompiling with this patch applied.


LoupVaillant:
Oh, thanks a lot. Merged. I've taken the opportunity to put the soname in a variable, to help with our upcoming ABI break.

PR #137, opened 2019-12-05
Man page updates

fscoto:
Mainly fixes for the new API, but also some other minor things that bothered me so much while reading them that I just went and fixed them as well while already there.

Also includes a change to the makefile and `man2html.sh`. Not 100% sure about optional/ in man yet, but I think it's reasonable since it's flattened at install time.


doc/man/man3/crypto_sign_init_first_pass_custom_hash.3monocypher
```diff
@@ -0,0 +1,234 @@
...
+.Em must
+be exactly 64 bytes.
+This will normally be constructed using the functions that provide the
+.Fa init ,
+.Fa update
+and
+.Fa final
+members.
+.It Fa init
+Function that initialises the hash context an outer signing context.
```
LoupVaillant: *of* an outer signing context?

doc/man/man3/crypto_sign_init_first_pass_custom_hash.3monocypher
```diff
@@ -0,0 +1,234 @@
...
+    crypto_ed25519_sign(theirs, sk, NULL, msg, sizeof(msg));
+
+    struct outer_ctx ctx;
+    crypto_sign_init_first_pass_custom_hash((void*)&ctx,
+                                            sk, NULL, &my_vtable);
+    crypto_sign_update(          (void*)&ctx, msg, sizeof(msg));
+    crypto_sign_init_second_pass((void*)&ctx);
+    crypto_sign_update(          (void*)&ctx, msg, sizeof(msg));
+    crypto_sign_final(           (void*)&ctx, mine);
+
```
LoupVaillant: You may want to wipe the secret key, unless you think that'd clutter the example (the context itself is wiped, so no need to do that explicitly).

doc/man/man3/intro.3monocypher
```diff
@@ -68,16 +90,13 @@ the chances of leaks.
...
 .Xr crypto_blake2b_init 3monocypher ,
 .Xr crypto_blake2b_update 3monocypher ,
-.Xr crypto_chacha20_encrypt 3monocypher ,
-.Xr crypto_chacha20_H 3monocypher ,
-.Xr crypto_chacha20_init 3monocypher ,
-.Xr crypto_chacha20_set_ctr 3monocypher ,
-.Xr crypto_chacha20_stream 3monocypher ,
-.Xr crypto_chacha20_x_init 3monocypher ,
+.Xr crypto_chacha20 3monocypher ,
+.Xr crypto_chacha20_ctr 3monocypher ,
```
LoupVaillant: I guess you are omitting IETF Chacha20 on purpose? (Your call, I'm okay with that.)

doc/man/man3/optional/crypto_hmac_sha512.3monocypher
```diff
@@ -0,0 +1,162 @@
...
+is a convenience function that
+performs
+.Fn crypto_hmac_sha512_init ,
+.Fn crypto_hmac_sha512_update ,
+and
+.Fn crypto_hmac_sha512_final .
+.Pp
+Use
+.Xr crypto_verify64 3monocypher
+to compare MACs.
```
LoupVaillant: Actually, the main use of HMAC is the same as Poly1305: authenticating messages. For this, a 16 byte hash is more than enough. We can truncate the output to 16 bytes, and use `crypto_verify16(3)` to compare the MACs. It may or may not be worth laying out explicitly.

If you're using it to implement something like Noise however, you probably need the full output.

makefile
```diff
@@ -44,7 +44,10 @@ install: library src/monocypher.h install-doc
 
 install-doc:
 	mkdir -p $(MAN_DIR)
-	cp -r doc/man/man3/*.3monocypher $(MAN_DIR)
+	cp -PR doc/man/man3/*.3monocypher $(MAN_DIR)
+ifdef USE_ED25519
+	cp -PR doc/man/man3/optional/*.3monocypher $(MAN_DIR)
+endif
```
LoupVaillant: :+1:

PR #141, opened 2019-12-12
Revert "intro: format link with .Lk"

fscoto:
This reverts commit df321b37cda1db5e8bd6fa76e4daf01ac087b865.

As it turns out,
groff (1.22.4) will spew complaints about the Lk formatting to stderr,
even though it's rendered perfectly fine.


LoupVaillant:
Ah, too bad. It would have helped HTML generation. Bah, it doesn't really matter. Merging.

PR #146, opened 2019-12-28
Rewrite example test script in shell

mikejsavage:
Still kinda busted with the custom hash example


fscoto:
As for the more pressing issue about the custom hash example – I think the easiest solution would be to have a special case for it and skip that page in particular since there's no reasonable way to do it inside an existing main() function as functions cannot be declared inside functions.

mikejsavage:
Thanks, made some changes

That custom hash API doesn't seem like it fits in with the rest of monocypher. E.g. monocypher doesn't have the argon2/blake2 variants

LoupVaillant:
> That custom hash API doesn't seem like it fits in with the rest of monocypher. E.g. monocypher doesn't have the argon2/blake2 variants

Argon2 doesn't accept custom hashes because (i) baseline Monocypher provides Blake2b, and (ii) nobody ever asked me to add it. Orthogonality isn't worth the extra complication.

And yes, some custom hash functions are weird, and not quite orthogonal. I did this mostly to preserve the old API (Monocypher 3 manages to *never* break the signature API), and simplicity (hijacking the existing incremental API allowed me to limit the number of functions).

doc/extract_examples.sh
```diff
@@ -71,7 +71,7 @@ for f in man/man3/*.3monocypher man/man3/optional/*.3monocypher; do
 	# and can't be tested like this
 	if [ ! -h "$f" ] && [ "$f" != "man/man3/crypto_sign_init_first_pass_custom_hash.3monocypher" ]; then
 		echo "// $f"
-		sed -n "/^\.Bd/,/^\.Ed/p" < $f | sed "s/\.Bd.*/{/" | sed "s/\.Ed/}/"
+		cat "$f" | sed -n "/^\.Bd/,/^\.Ed/p" | sed "s/\.Bd.*/{/" | sed "s/\.Ed/}/"
```
fscoto: Useless use of cat? If you want just left to right, you could've moved `<$f` to the beginning of the line instead. Not sure if you were aware of this.

doc/extract_examples.sh
```diff
@@ -77,4 +77,5 @@ do
     fi
 done
 
+echo "return 0;"
```
fscoto: `main`() always returns 0 implicitly; C11 § 5.1.2.2.3(1):

> If the return type of the `main` function is a type compatible with `int`, [...] reaching the `}` that terminates the `main` function returns a value of 0.

Not strictly necessary to add this, though it's probably good style to.

doc/extract_examples.sh
```diff
@@ -0,0 +1,78 @@
...
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+# ------------------------------------------------------------------------
+#
+# Written in 2017 and 2019 by Loup Vaillant and Fabio Scotoni
```
fscoto: *Michael Savage

doc/extract_examples.sh
mikejsavage: Useless cat but easier to understand

I don't think I've ever seen that syntax before so I'll leave it as cat

LoupVaillant:
Back from vacation at last.

Sorry for the delays, let the merges begin.

And have a happy new year. :-)

PR #147, opened 2019-12-28
Some examples fixes

mikejsavage:
Lots of the examples don't compile as C++ because they call crypto_wipe on const buffers

Having consts makes it a little clearer what's happening in the function calls that actually do the thing and these examples won't work by copy pasting them because they don't provide any data so IMO it's ok. It's not totally consistent right now so either way more changes are needed

### DIFF
--- a/doc/man/man3/crypto_chacha20.3monocypher
+++ b/doc/man/man3/crypto_chacha20.3monocypher
@@ -243,7 +243,7 @@ const uint8_t key        [ 32]; /* Secret random key              */
 const uint8_t nonce      [ 24]; /* Unique nonce (possibly random) */
 const uint8_t plain_text [500]; /* Message to be encrypted        */
 uint8_t       cipher_text[500]; /* Will be the encrypted message  */
-crypto_xchacha20_encrypt(cipher_text, plain_text, 500, key, nonce);
+crypto_xchacha20(cipher_text, plain_text, 500, key, nonce);
 /* Wipe secrets if they are no longer needed */
 crypto_wipe(key,        32);
 crypto_wipe(plain_text, 500);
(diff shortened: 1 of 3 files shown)


doc/man/man3/crypto_chacha20.3monocypher
```diff
@@ -269,15 +269,16 @@ const uint8_t nonce      [ 24]; /* Unique nonce (possibly random) */
 const uint8_t plain_text [500]; /* Message to be encrypted        */
 uint8_t       cipher_text[500]; /* Will be the encrypted message  */
 uint64_t      ctr;              /* Block counter */
-for(int i = 0; i < 500; i += 64) {
-    ctr = crypto_xchacha20_ctr(&ctx, cipher_text+i, plain_text+i,
-                               64, ctr);
+int i;
+for(i = 0; i < 500; i += 64) {
```
fscoto: Monocypher requires a C99 compiler since it uses `stdint.h` and C99-style for loops. If the build failed because of/warned about `for(int i = 0; /* ... */)` being a C99 feature, that's a non-issue and needs to be worked around with compiler flags instead.

mikejsavage: i gets used after the loop

fscoto: Ah, my bad with `int i;`. You're right.

This code is tricky. At `i == 448`, 64 bytes will be written, overflowing the buffer, oh dear.
I *think* the for loop was actually supposed to have a condition like `i < (500 - 64)`.
In fact, if we follow your idea with min() – though that'll have to be a ternary operation instead of min() since min() isn't actually a thing in the C stdlib –, the part after the for loop shouldn't even be necessary anymore since the loop properly takes care of it.

PR #151, opened 2020-01-15
Add some comments about what the EC functions do

fscoto:
This hopefully helps both auditing the code (by giving a clear
indication of what a function is supposed to be doing),
and trying to work with it (by minimizing the amount of time people need
to determine if a function is relevant to their interests).

---

This is basically just an assortment of things I wish I knew these before going into the code. I'm not entirely sure that `ge` stands for "group element", but that's what I suspect it means.

While already here, remove two spaces in front of semicolons.

### DIFF
--- a/src/monocypher.c
+++ b/src/monocypher.c
@@ -1127,6 +1127,7 @@ static void fe_frombytes(fe h, const u8 s[32])
     FE_CARRY;
 }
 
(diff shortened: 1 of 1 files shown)


LoupVaillant:
This patch is perfect. As I suspected much better than I would have done. Merging right away, thanks.

Issue #154, opened 2020-02-05
Trouble understanding and using Monocypher


aggsol:
I am trying to use Monocypher to hash user passwords. While I can read the manual front page
all links to functions like `crypto_blake2b()` of `crypto_blake2b_final()` are not working. I can of course read the header and see how e. g. the Blake2 functions should be called.

But even then it s not clear to me how to add the salt to the hash. Do I hash the salt first or last? Should I append or prepend the salt to the password? How can I hash with multiple rounds?

In the current state I don't feel that I am using the library correctly. I fear I do some stupid error and the security is broken already. Better documentation or examples would provide confidence for non-crypto programmers that need to use it.

fscoto:
> While I can read the manual front page
all links to functions like `crypto_blake2b()` of `crypto_blake2b_final()` are not working.

Oh dear. That's an issue caused by Loup's workflow for the web manual and probably needs urgent fixing if you're running into this as an issue. As a workaround, you can use:

- the sidebar on the website rather than trying to click on the (currently fake) links on the intro page of the web manual,
- the man pages (`make install-doc` -> `man 3monocypher intro`, `man crypto_blake2b`, etc.) if on you use \*NIX,
- my separate [man-ocypher](https://fscoto.github.io/man-ocypher/intro.html) (which uses a different man page -> web translation), where cross-page links definitely work (but there's no sidebar).

---

> But even then it s not clear to me how to add the salt to the hash. Do I hash the salt first or last? Should I append or prepend the salt to the password? How can I hash with multiple rounds?

None of these, actually. BLAKE2b is a general-purpose cryptographic hash. You're not supposed to sue the `crypto_blake2b_*` family of functions for password hashing at all. Password hashing should use a specialized, computation-intensive hashing function to delay attackers from churning through a lot of hashes very quickly (which is, I presume, why you asked about multiple rounds).

You're looking for "password key derivation", i.e. the [`crypto_argon2i()` functions](https://monocypher.org/manual/argon2i). Password hashing and password key derivation happen to have the same requirements; in the case of password hashing for storing passwords, you store the derived key (== the password hash) as such instead of using it for encryption operations. Perhaps that should've been stated more clearly in the intro page.

I hope that helps; if you have any follow-up questions, feel free to ask.

aggsol:
[...]

PR #155, opened 2020-02-05
Be more explicit about passwords in man pages

fscoto:
Prompted by an inquiry in #154.

---

The manual work this time is unusually deliberate in multiple ways.

**BLAKE2b/SHA-512**: The wording "BLAKE2b itself is not suitable" is a compromise between saying "*BLAKE2b is not suitable*" (which is an assertion I'm not comfortable making while PBKDF2 is still a thing that exists and can work with BLAKE2b as well as anything else—though one may certainly argue how appropriate PBKDF2 still is in times we bicker about Argon2 not having cache resistance) and "*BLAKE2b by itself is not suitable*" (which would correctly imply there are constructions that may be suitable, but then the question becomes "Why doesn't Monocypher just do this?", and that's one that would go far beyond what the manual is supposed to answer). I'm still not 100% happy with the wording I chose here, but it's an inconvenient tightrope either way.

**Security considerations**: I did pick up your idea from #154 to have a "Security Considerations" section, but I'm also okay with it being removed. You can't really talk about the issue at length because doing so would end up reinventing applied cryptography books while ultimately not providing much for the actual usage of the functions. While the SHA-512 page mostly refers to the BLAKE2b page, I've intentionally mirrored the "Security Considerations" section here in case people only selectively read the BLAKE2b page.

**Intro**: Mention password hashing right next to password key derivation. This should help prevent confusion about whether you want a "hash" function or a "password key derivation" function for passwords; the answer is always the latter. This slightly stretches the meaning of the original wording to go beyond what used to be since the relationship between password key derivation and password hashing is a non-trivial one, but this change helps the uninitiated and does no harm for the experienced reader. This may need a corresponding change in the sidebar of the manual on the website.

Authorship notices and mdoc dates bumped as required.

### DIFF
--- a/doc/man/man3/crypto_blake2b.3monocypher
+++ b/doc/man/man3/crypto_blake2b.3monocypher
@@ -10,7 +10,7 @@
 .\"
 .\" Copyright (c) 2017-2019 Loup Vaillant
 .\" Copyright (c) 2018 Michael Savage
-.\" Copyright (c) 2017, 2019 Fabio Scotoni
+.\" Copyright (c) 2017, 2020 Fabio Scotoni
 .\" All rights reserved.
 .\"
 .\"
@@ -40,7 +40,7 @@
 .\"
 .\" ----------------------------------------------------------------------------
 .\"
-.\" Written in 2017-2019 by Loup Vaillant, Michael Savage and Fabio Scotoni
+.\" Written in 2017-2020 by Loup Vaillant, Michael Savage and Fabio Scotoni
 .\"
 .\" To the extent possible under law, the author(s) have dedicated all copyright
 .\" and related neighboring rights to this software to the public domain
@@ -50,7 +50,7 @@
 .\" with this software.  If not, see
 .\" <https://creativecommons.org/publicdomain/zero/1.0/>
 .\"
-.Dd December 12, 2019
+.Dd February 5, 2020
 .Dt CRYPTO_BLAKE2B 3MONOCYPHER
 .Os
 .Sh NAME
@@ -104,6 +104,11 @@
 BLAKE2b is a fast cryptographically secure hash, based on the ideas of
 Chacha20.
 It is faster than MD5, yet just as secure as SHA-3.
+However, BLAKE2b itself
+.Sy is not suitable for hashing and deriving keys from passwords ;
+use the
+.Xr crypto_argon2i 3monocypher
+family of functions for that purpose instead.
 .Pp
 BLAKE2b is immune to length extension attacks, and as such does not
 require any specific precautions, such as using the HMAC algorithm.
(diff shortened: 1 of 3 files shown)


doc/man/man3/crypto_blake2b.3monocypher
```diff
@@ -104,6 +104,11 @@
 BLAKE2b is a fast cryptographically secure hash, based on the ideas of
 Chacha20.
 It is faster than MD5, yet just as secure as SHA-3.
+However, BLAKE2b itself
+.Sy is not suitable for hashing and deriving keys from passwords ;
```
aggsol: This sounds like Blake2b is not suitable for general hashing while you actually mean "not suitable for hashing passwords and deriving key from passwords"

doc/man/man3/crypto_blake2b.3monocypher
```diff
@@ -276,3 +281,18 @@ Any deviation from the specified input and output length ranges results
 in
 .Sy undefined behaviour .
 Make sure your inputs are correct.
+.Sh SECURITY CONSIDERATIONS
+BLAKE2b is a general-purpose cryptographic hash function;
+this means that it is not suited for hashing passwords and deriving
```
aggsol: For the average coder it would be helpful to state first some example what Blake3b is *suitable* for before telling what it is *not suitable* for. Just for a clear separation.

doc/man/man3/crypto_blake2b.3monocypher
fscoto: Point given. I was trying to be brief, but that was probably abbreviating a bit much.

doc/man/man3/crypto_blake2b.3monocypher
fscoto: I'm not really sure how to approach this issue, both from a "How do I avoid writing a book on applied cryptography in the form of a sequence of man pages?" perspective, and from a "What is actually useful for most people to know?" The problem with explaining that is that there is too much it *is* useful for. Applications include:

- checking if data is identical (e.g. quick file sameness comparison)
- checking if data is unchanged from some known-good value (though usually you'll want signtures or MACs instead), see also hash trees for doing this at scale
- *non-password* key derivation (as required in some protocols, IIRC PURBs make use of this)
- as a core component of signatures (but Monocypher figures that out for you as part of `crypto_sign()`, so you'll never actually have to touch this yourself)
- as part of random number generators (but it seems everyone's moved to stream ciphers instead)
- message authentication codes, i.e. symmetric signatures (BLAKE2b in keyed mode; SHA-512 with HMAC); though the manual actually mentions this one when talking about the `key` argument
- as part of PBKDF2 (but please don't; Monocypher has Argon2)
- as part of key agreement (hashing in the conversation until a certain point, cf. Monokex)
- ...

Most of these are only necessary for implementing more elaborate constructions yourself and I feel like this list just creates more questions than it answers. I'm not sure what you're looking for here. See also the [NaCl documentation, which categorizes hash functions as "low-level functions"](https://nacl.cr.yp.to/hash.html) instead.

CC @LoupVaillant since intro(3monocypher) also tells people to read Crypto101.

LoupVaillant:
That patch is perfect, thanks a lot.

Your _security considerations_ section works for me. It succinctly  explains why you don't want to use a regular hash for handling passwords, and is short enough that it doesn't get in the way. I say we keep it.

@aggsol, I disagree however with expanding hash uses (and non uses) any further. Going through examples would mainly make the manual more cumbersome. And as much as it sucks, those man pages are a reference manual. They can teach you how to use Monocypher, but they are not intended to teach applied cryptography. Users are supposed to have some basic understanding before they can even hope to safely use any cryptographic library of any kind.

We may someday write an applied cryptography tutorial, focused on using modern crypto libraries such as Monocypher or Libsodium. That will be a separate project.

aggsol:
Yeah, I totally understand your arguments here. Thank you and keep up the good work!

PR #158, opened 2020-02-29
Add random_bytes() to crypto_lock examples

fscoto:
I'm just doing this to crypto_lock.3monocypher right now to see if this is an improvement.

While already there, I've taken a few other liberties for clarification (use `else` with `crypto_unlock` because there's not enough outer structure to suggest just `return -1` with no cleanup or `goto clean` so that it doesn't look like it's dangling into the decryption-good part anyway; there are contexts other than sending stuff over the network).

So far, I'm not really convinced of any of those changes though and do mildly believe that the status quo is better:

- The `random_bytes` introduction is kind of wordy and just triplicates the usual hint to generate randomness with the OS (once in the intro, once in the description for the `nonce` argument, once in the new paragraph).
- The additional context comments and `if`/`else` stuff is complex to understand—primarily because it is abstract. An abstract example is of only limited help, however, so I'm thinking maybe we'd be better off saying “we're assuming standard UNIX networking functions” and having a single example program talk to itself over a socketpair(2) with a random key, where the encryption part uses split buffers and the decryption part uses in-place decryption.

### DIFF
--- a/doc/man/man3/crypto_lock.3monocypher
+++ b/doc/man/man3/crypto_lock.3monocypher
@@ -10,7 +10,7 @@
 .\"
 .\" Copyright (c) 2017-2019 Loup Vaillant
 .\" Copyright (c) 2017-2018 Michael Savage
-.\" Copyright (c) 2017, 2019 Fabio Scotoni
+.\" Copyright (c) 2017, 2019-2020 Fabio Scotoni
 .\" All rights reserved.
 .\"
 .\"
@@ -40,7 +40,7 @@
 .\"
 .\" ----------------------------------------------------------------------------
 .\"
-.\" Written in 2017-2019 by Loup Vaillant, Michael Savage and Fabio Scotoni
+.\" Written in 2017-2020 by Loup Vaillant, Michael Savage and Fabio Scotoni
 .\"
 .\" To the extent possible under law, the author(s) have dedicated all copyright
 .\" and related neighboring rights to this software to the public domain
@@ -50,7 +50,7 @@
 .\" with this software.  If not, see
 .\" <https://creativecommons.org/publicdomain/zero/1.0/>
 .\"
-.Dd December 12, 2019
+.Dd February 29, 2020
 .Dt CRYPTO_LOCK 3MONOCYPHER
 .Os
 .Sh NAME
@@ -226,6 +226,21 @@ attacker's interference.
 .Fa plain_text
 does not need to be wiped if the decryption fails.
 .Sh EXAMPLES
+The following examples assume that a function called
+.Fn random_bytes
+exists.
+It fills the given buffer with cryptographically secure random
+bytes
+(see
+.Xr intro 3monocypher
+for some advice on how to accomplish this yourself).
+The function has this prototype:
+.Ft void
+.Fo random_bytes
+.Fa "uint8_t *buf"
+.Fa "size_t len"
+.Fc
+.Pp
 Encryption:
 .Bd -literal -offset indent
 uint8_t       key        [32];  /* Random, secret session key  */
@@ -233,11 +248,14 @@ const uint8_t nonce      [24];  /* Use only once per key       */
 uint8_t       plain_text [500]; /* Secret message              */
 uint8_t       mac        [16];  /* Message authentication code */
 uint8_t       cipher_text[500]; /* Encrypted message           */
+random_bytes(key, 32);
 crypto_lock(mac, cipher_text, key, nonce, plain_text, 500);
 /* Wipe secrets if they are no longer needed */
 crypto_wipe(plain_text, 500);
 crypto_wipe(key, 32);
-/* Transmit cipher_text, nonce, and mac over the network */
+/* Transmit cipher_text, nonce, and mac over the network,
+ * store them in a file, etc.
+ */
 .Ed
 .Pp
 To decrypt the above:
@@ -253,10 +271,12 @@ if (crypto_unlock(plain_text, key, nonce, mac, cipher_text, 500)) {
      * and abort the decryption.
      */
     crypto_wipe(key, 32);
+} else {
+    /* ...do something with the decrypted text here... */
+    /* Finally, wipe secrets if they are no longer needed */
+    crypto_wipe(plain_text, 500);
+    crypto_wipe(key, 32);
 }
-/* Wipe secrets if they are no longer needed */
-crypto_wipe(plain_text, 500);
-crypto_wipe(key, 32);
 .Ed
 .Pp
 In-place encryption:
@@ -265,10 +285,13 @@ uint8_t       key  [32];  /* Random, secret session key  */
 const uint8_t nonce[24];  /* Use only once per key       */
 uint8_t       text [500]; /* Secret message              */
 uint8_t       mac  [16];  /* Message authentication code */
+random_bytes(key, 32);
 crypto_lock(mac, text, key, nonce, text, 500);
 /* Wipe secrets if they are no longer needed */
 crypto_wipe(key, 32);
-/* Transmit text, nonce, and mac over the network */
+/* Transmit cipher_text, nonce, and mac over the network,
+ * store them in a file, etc.
+ */
 .Ed
 .Pp
 In-place decryption:
@@ -283,10 +306,12 @@ if (crypto_unlock(text, key, nonce, mac, text, 500)) {
      * and abort the decryption.
      */
     crypto_wipe(key, 32);
+} else {
+    /* ...do something with the decrypted text here... */
+    /* Finally, wipe secrets if they are no longer needed */
+    crypto_wipe(text, 500);
+    crypto_wipe(key, 32);
 }
-/* Wipe secrets if they are no longer needed */
-crypto_wipe(text, 500);
-crypto_wipe(key, 32);
 .Ed
 .Sh SEE ALSO
 .Xr crypto_key_exchange 3monocypher ,
--- a/doc/man/man3/crypto_lock.3monocypher
+++ b/doc/man/man3/crypto_lock.3monocypher
@@ -226,32 +226,28 @@ attacker's interference.
 .Fa plain_text
 does not need to be wiped if the decryption fails.
 .Sh EXAMPLES
-The following examples assume that a function called
-.Fn random_bytes
-exists.
-It fills the given buffer with cryptographically secure random
-bytes
-(see
+The following examples assume the existence of
+.Fn arc4random_buf ,
+which fills the given buffer with cryptographically secure random bytes.
+If
+.Fn arc4random_buf
+does not exist on your system, see
 .Xr intro 3monocypher
-for some advice on how to accomplish this yourself).
-The function has this prototype:
-.Ft void
-.Fo random_bytes
-.Fa "uint8_t *buf"
-.Fa "size_t len"
-.Fc
+for advice about how to generate cryptographically secure random bytes.
 .Pp
 Encryption:
 .Bd -literal -offset indent
-uint8_t       key        [32];  /* Random, secret session key  */
-const uint8_t nonce      [24];  /* Use only once per key       */
-uint8_t       plain_text [500]; /* Secret message              */
-uint8_t       mac        [16];  /* Message authentication code */
-uint8_t       cipher_text[500]; /* Encrypted message           */
-random_bytes(key, 32);
-crypto_lock(mac, cipher_text, key, nonce, plain_text, 500);
+uint8_t key        [32];    /* Random, secret session key  */
+uint8_t nonce      [24];    /* Use only once per key       */
+uint8_t plain_text [12] = "Lorem ipsum"; /* Secret message */
+uint8_t mac        [16];    /* Message authentication code */
+uint8_t cipher_text[12];              /* Encrypted message */
+arc4random_buf(key,   32);
+arc4random_buf(nonce, 24);
+crypto_lock(mac, cipher_text, key, nonce, plain_text,
+        sizeof(plain_text));
 /* Wipe secrets if they are no longer needed */
-crypto_wipe(plain_text, 500);
+crypto_wipe(plain_text, 12);
 crypto_wipe(key, 32);
 /* Transmit cipher_text, nonce, and mac over the network,
  * store them in a file, etc.
@@ -260,12 +256,12 @@ crypto_wipe(key, 32);
 .Pp
 To decrypt the above:
 .Bd -literal -offset indent
-uint8_t       key        [32];  /* Same as the above         */
-const uint8_t nonce      [24];  /* Same as the above         */
-const uint8_t cipher_text[500]; /* Encrypted message         */
-const uint8_t mac        [16];  /* Received from the network */
-uint8_t       plain_text [500]; /* Secret message            */
-if (crypto_unlock(plain_text, key, nonce, mac, cipher_text, 500)) {
+uint8_t       key        [32]; /* Same as the above        */
+uint8_t       nonce      [24]; /* Same as the above        */
+const uint8_t cipher_text[12]; /* Encrypted message        */
+const uint8_t mac        [16]; /* Received along with text */
+uint8_t       plain_text [12]; /* Secret message           */
+if (crypto_unlock(plain_text, key, nonce, mac, cipher_text, 12)) {
     /* The message is corrupted.
      * Wipe key if it is no longer needed,
      * and abort the decryption.
@@ -274,19 +270,20 @@ if (crypto_unlock(plain_text, key, nonce, mac, cipher_text, 500)) {
 } else {
     /* ...do something with the decrypted text here... */
     /* Finally, wipe secrets if they are no longer needed */
-    crypto_wipe(plain_text, 500);
+    crypto_wipe(plain_text, 12);
     crypto_wipe(key, 32);
 }
 .Ed
 .Pp
 In-place encryption:
 .Bd -literal -offset indent
-uint8_t       key  [32];  /* Random, secret session key  */
-const uint8_t nonce[24];  /* Use only once per key       */
-uint8_t       text [500]; /* Secret message              */
-uint8_t       mac  [16];  /* Message authentication code */
-random_bytes(key, 32);
-crypto_lock(mac, text, key, nonce, text, 500);
+uint8_t key  [32];    /* Random, secret session key  */
+uint8_t nonce[24];    /* Use only once per key       */
+uint8_t text [12] = "Lorem ipsum"; /* Secret message */
+uint8_t mac  [16];    /* Message authentication code */
+arc4random_buf(key,   32);
+arc4random_buf(nonce, 24);
+crypto_lock(mac, text, key, nonce, text, 12);
 /* Wipe secrets if they are no longer needed */
 crypto_wipe(key, 32);
 /* Transmit cipher_text, nonce, and mac over the network,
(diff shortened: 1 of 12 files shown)


doc/man/man3/crypto_lock.3monocypher
```diff
@@ -226,18 +226,36 @@ attacker's interference.
...
+(see
+.Xr intro 3monocypher
+for some advice on how to accomplish this yourself).
+The function has this prototype:
+.Ft void
+.Fo random_bytes
+.Fa "uint8_t *buf"
+.Fa "size_t len"
+.Fc
+.Pp
```
LoupVaillant: I believe those are overkill, and best omitted. Instead, a short *code* comment in the example itself should do:

    random_bytes(key, 32) /* use the OS to implement this function */

Alternatively, we could just give up and use arc4random_buf(3), which is my recommended API whenever available anyway.

doc/man/man3/crypto_lock.3monocypher
```diff
@@ -226,18 +226,36 @@ attacker's interference.
...
 Encryption:
 .Bd -literal -offset indent
 uint8_t       key        [32];  /* Random, secret session key  */
 const uint8_t nonce      [24];  /* Use only once per key       */
 uint8_t       plain_text [500]; /* Secret message              */
 uint8_t       mac        [16];  /* Message authentication code */
 uint8_t       cipher_text[500]; /* Encrypted message           */
+random_bytes(key, 32);
```
LoupVaillant: So, this one could be replaced by one of the following:

    random_bytes(key, 32) /* use the OS to implement this function */
    arc4random_buf(key, 32);

If we're starting to use standard UNIX functions, I believe I prefer arc4random_buf(3), which requires less extraneous explanation (possibly an `#include` line, no more). Plus, it will follow our own advice and use the damn OS to get random stuff.

doc/man/man3/crypto_lock.3monocypher
```diff
@@ -253,10 +271,12 @@ if (crypto_unlock(plain_text, key, nonce, mac, cipher_text, 500)) {
      * and abort the decryption.
      */
     crypto_wipe(key, 32);
+} else {
+    /* ...do something with the decrypted text here... */
+    /* Finally, wipe secrets if they are no longer needed */
+    crypto_wipe(plain_text, 500);
+    crypto_wipe(key, 32);
```
LoupVaillant: I agree with that part.

doc/man/man3/crypto_lock.3monocypher
```diff
@@ -226,18 +226,36 @@ attacker's interference.
...
 uint8_t       cipher_text[500]; /* Encrypted message           */
+random_bytes(key, 32);
 crypto_lock(mac, cipher_text, key, nonce, plain_text, 500);
 /* Wipe secrets if they are no longer needed */
 crypto_wipe(plain_text, 500);
 crypto_wipe(key, 32);
-/* Transmit cipher_text, nonce, and mac over the network */
+/* Transmit cipher_text, nonce, and mac over the network,
+ * store them in a file, etc.
+ */
```
LoupVaillant: Sockets would work, but I think reading and writing to the standard input & output is better, because it will work on other systems. That way we're closer to code working out of the box. (Even real programs sometimes use their standard input & output, which are then redirected to the network.)

Alternatively, just use some unspecified incoming and outgoing (or input and output, or whatever pair of name you think is best) channels, and use read(2) and write(2) on them. Or fread(3) and fwrite(3) if they're easier to use.

doc/man/man3/crypto_lock.3monocypher
fscoto: > Alternatively, we could just give up and use arc4random_buf(3), which is my recommended API whenever available anyway.

As much as I would love to do that (and have done that for the custom hash function documentation because I was assuming the platform as a whole to have an alternative SHA-2 implementation to test with), I don't think that's a good idea for the *general-purpose* documentation. Please recall that the dominant systems are Windows, macOS and Linux, of which only macOS provides arc4random_buf(3). Ignoring the embedded non-Linux case (bare metal, custom OS, VxWorks come to mind), arc4random_buf(3) is provided by niche operating systems only ({Free,Open,Net}BSD, illumos, Solaris). So in the most common cases, there's a good chance that people will run into a snag.

Said snag is also a “hard” problem: They will be faced with a compilation error for arc4random_buf(3) being missing, rather than a more obviously named random_bytes() function. The former might make them look into all kinds of wrong things (“Am I missing something?”) rather than re-checking the comment (ideally) or googling a more generically-useful query that hopefully leads to a StackOverflow post that is sufficiently generic. I admit this is speculative, but I'd rather avoid having a *specific* problem that stands in for misunderstanding a *generic* problem.

I do think I'll roll with the comment instead of arc4random() and take out the paragraph to keep it concise.

fscoto:
I think a couple of things here depend on whether we should just assume *NIX/POSIX interfaces or would like to avoid doing so.

Setting up any kind of program that communicates with itself is an issue. We can use fread and fwrite as guaranteed by the C standard itself to have it communicate over a file and would thus need a file-like scenario. socketpair(2) or just plain socket(2) is POSIX-specific again (though Windows tries to approximate those in Winsock as far as I can tell). So I believe it'd be best if we make up our minds on whether to assume POSIX and make a non-trivial examples, leave it as-is, just go with something like `random_bytes(key, 32); /* use OS facilities to implement this */` or cutting the trivial examples entirely.

doc/man/man3/crypto_lock.3monocypher
LoupVaillant: My Ubuntu seems to provide arc4random(3). At least there's the man page. Windows users are indeed left in the dust.

> Said snag is also a “hard” problem: They will be faced with a compilation error for arc4random_buf(3) being missing, rather than a more obviously named random_bytes() function. 

Ah, that's a problem indeed. `random_bytes()` it is, then.

> I do think I'll roll with the comment instead of arc4random() and take out the paragraph to keep it concise.

Works for me.

LoupVaillant:
What to do with the example is tricky. I think the examples should be kept, because I believe many users work primarily with them (I don't). At the same time, adding *anything* that does not pertain to the documented function directly would just add clutter for no good reason.

My (weak) opinion right now is that we should provide mock functions for everything: random, read, write. We could begin all examples by a little comment like this:

    /* The functions below are placeholders:                       */
    /* random_bytes() generates random bytes                       */
    /* read_bytes() reads bytes from some input  (file, socket...) */
    /* write_bytes() writes bytes to some output (file, socket...) */

doc/man/man3/crypto_lock.3monocypher
fscoto: > My Ubuntu seems to provide arc4random(3).

For the sake of completeness:

It provides libbsd as part of the libbsd0 package due to dependencies of other packages that were installed by default (`apt-cache rdepends --installed libbsd0`, which for your machine will probably include xserver-xorg-core, libedit2 [via openssh-client] and maybe samba). Using libbsd is a bit complicated; looking at libbsd(7), you either prefix your includes like so: `#include <bsd/unistd.h>` (which breaks on the BSDs themselves, or you use it in "overlay mode", compiling with `-isystem /usr/include/bsd -DLIBBSD_OVERLAY` (assuming the installation prefix was `/usr`, cf. `pkg-config --cflags libbsd-overlay`) and linking with `-lbsd` at the end in either case. You cannot rely on libbsd existing on all distributions, however; it'll be routinely missing from e.g. Alpine Linux.

Here's a handy table summarizing what I know at the top of my head:

Mechanism | OS | Notes
-|-|-
getrandom(2) | Linux, FreeBSD, illumos, Solaris | recent Linux invention; illumos tells you to prefer arc4random; fairly recent invention; can fail; OpenBSD resists adoption
getentropy() | OpenBSD, Linux, FreeBSD, NetBSD, illumos, Solaris | older OpenBSD invention; only supposed to be used to *seed* an RNG (used by arc4random internally); on standardization track for POSIX; emulated on Linux in glibc and is hence getentropy(*3*) there; limit of 256 bytes per invocation; can fail
arc4random(3) | every BSD, macOS, illumos, Solaris | OpenBSD invention, kernel-seeded userspace PRNG that reseeds itself periodically (which is the reason getentropy(2) was invented in the first place, since it originally used `/dev/urandom`); early versions used actual RC4 and had statistical bias accordingly (still affects FreeBSD 11 [!] but not 12); does not fail and instead aborts the program if getting entropy from kernel fails (the kernel is supposed to guarantee delivery of entropy as part of the contract with the arc4random interface); part of libc, not the kernel interface
`/dev/urandom` | almost every *NIX like including stuff like HP-UX | can fail in a myriad of ways and be missing in a chroot, but at least it's available almost everywhere
BCryptGenRandom() | Windows | The way you're supposed to do it on Windows; can actually operate without an algorithm handle apparently (unlike the XP-compatible CryptGenRandom); can fail
RtlGenRandom() | Windows | The way libsodium does it (whose comments say so do Rust, Firefox and BoringSSL)

fscoto: > My Ubuntu seems to provide arc4random(3).

For the sake of completeness:

It provides libbsd as part of the libbsd0 package due to dependencies of other packages that were installed by default (`apt-cache rdepends --installed libbsd0`, which for your machine will probably include xserver-xorg-core, libedit2 [via openssh-client] and maybe samba). Using libbsd is a bit complicated; looking at libbsd(7), you either prefix your includes like so: `#include <bsd/unistd.h>` (which breaks on the BSDs themselves, or you use it in "overlay mode", compiling with `-isystem /usr/include/bsd -DLIBBSD_OVERLAY` (assuming the installation prefix was `/usr`, cf. `pkg-config --cflags libbsd-overlay`) and linking with `-lbsd` at the end in either case. You cannot rely on libbsd existing on all distributions, however; it'll be routinely missing from e.g. Alpine Linux.

Here's a handy table summarizing what I know at the top of my head:

Mechanism | OS | Notes
-|-|-
getrandom(2) | Linux, FreeBSD, illumos, Solaris | recent Linux invention; illumos tells you to prefer arc4random; fairly recent invention; can fail; OpenBSD resists adoption
getentropy() | OpenBSD, Linux, FreeBSD, NetBSD, illumos, Solaris | older OpenBSD invention; only supposed to be used to *seed* an RNG (used by arc4random internally); on standardization track for POSIX; emulated on Linux in glibc and is hence getentropy(*3*) there; limit of 256 bytes per invocation; can fail
arc4random(3) | every BSD, macOS, illumos, Solaris | OpenBSD invention, kernel-seeded userspace PRNG that reseeds itself periodically (which is the reason getentropy(2) was invented in the first place, since it originally used `/dev/urandom`); early versions used actual RC4 and had statistical bias accordingly (still affects FreeBSD 11 [!] but not 12); does not fail and instead aborts the program if getting entropy from kernel fails (the kernel is supposed to guarantee delivery of entropy as part of the contract with the arc4random interface); part of libc, not the kernel interface
`/dev/urandom` | almost every *NIX like including stuff like HP-UX | can fail in a myriad of ways and be missing in a chroot, but at least it's available almost everywhere
BCryptGenRandom() | Windows | The way you're supposed to do it on Windows; can actually operate without an algorithm handle apparently (unlike the XP-compatible CryptGenRandom); can fail
RtlGenRandom() | Windows | The way libsodium does it (whose comments say so do Rust, Firefox and BoringSSL)

fscoto:
Do you mean having those four lines for every individual example, even in pages with multiple examples, or only for the first example per page? Mock options options are unsatisfactory regardless of their placement:

1. If they're detailed in prose, we've already established that your reaction is “those are overkill”.
2. If they're detailed as one comment per example, then that's +4 lines for every example, some of which barely span that much actual code themselves.
3. If they're detailed only once per example section, then there's a chance readers would miss it if the first example doesn't match their usage scenario.

LoupVaillant:
> Do you mean having those four lines for every individual example, even in pages with multiple examples, or only for the first example per page?

On top of the first example, with a blank line to separate that comment from the example itself.

> If they're detailed in prose, we've already established that your reaction is “those are overkill”.

Well, I may have overreacted to the amount of additional source code. It should be much more compact on display. Thinking back, I'm actually okay with such a paragraph, but I believe it can be shorter (comparable in length to the comment header I wrote, perhaps a bit longer).

My *new* current opinion is that a short paragraph just below the EXAMPLE was the right idea after all. I just want it as short as possible. What you wrote felt a bit long. I think you can either omit the prototype entirely. Something like this, perhaps:

    The following examples assume the existence of a
    .Fn random_bytes
    function, that fills the given buffer with cryptographically secure
    random bytes (see
    .Xr intro 3monocypher
    for advice about how to do it.).
    .Pp

The idea is to cut as many words as I can. I believe the prototype is not needed, once you have the description of the function (it's a buffer and a length, how it works is pretty obvious if you have basic proficiency in C). Plus, this is just a placeholder. No need to document it as thoroughly as the public API.

jsfierro:
While I understand the concern about compiler warnings for people who don't have ar4random,  I prefer using  arc4random_buf() in the example to a mock function. Yes, a comment saying  "Use the CSPRNG supplied by your OS" is supposed to steer people towards the right path, but they might just call the non cryptographically secure random(). It's supplied by their OS, after all. An example with arc4random gets people closer to doing randomness the right way, even if they end up having to install libbsd or something to get it. 

If we're choosing between an example that may throw a compiler warning and an example that might lead to someone picking an insecure RNG, I'm strongly in favor of the former.  So much is broken due to poor quality random numbers  that anything that gets people closer to doing it right is worth doing IMHO.

LoupVaillant:
Good point. If we do this, it might be a good idea to keep the paragraph describing what arc4random(3) does, effectively treating it kind like a placeholder. What do you think, @fscoto?

fscoto:
I'll admit that I've got serious tunnel vision when it comes to RNGs since I've been deeply dissatisfied with the situation of RNGs in C and operating systems for years now, so the idea that someone might pick rand(3) or random(3) instead didn't even cross my mind, and that's my bad. arc4random_buf(3) now sounds a lot more reasonable.

How does this sound?

```mdoc
.Sh EXAMPLES
The following examples assume the existence of
.Fn arc4random_buf ,
which fills the given buffer with cryptographically secure random bytes.
If
.Fn arc4random_buf
does not exist on your system, see
.Xr intro 3monocypher
for advice about how to generate cryptographically secure random bytes.
.Pp
```

If we're randomizing the key, should we also randomize the nonce while we're already there? I figure we might as well; while implicit counting nonces are also an option (and might need demonstration—incrementing an array of 24 bytes is kind of its own exercise depending on how you do it), I figure the examples should do the safe default and just sacrifice the 24 bytes of bandwidth.

---

Incidentally, this still leaves up for debate what to do with the communication:

1. file and C stdio;
2. UNIX syscalls over a socketpair(2); or
3. add more boilerplate to the intro to detail imaginary `read_bytes` and `write_bytes` functions, then go in a similar direction as option 2;
4. keep the status quo and have separate, trivial examples.

Personally, I'm leaning towards option 4 here because option 1, full-file encryption, gets dangerously close to ransomware (perhaps this fear is unjustified, however), option 2 makes extra OS assumptions though they're probably justified by the point you have a working arc4random_buf(3) and option 3 we've just discarded for `random_bytes` so mixing approaches seems weird.

LoupVaillant:
Your boilerplate paragraph sounds good. One nitpick: perhaps use `.Xr arc4random_buf 3`? Might not be on the system, but if it is, it does have its manual section.

Randomizing the nonce sounds good. The nonce may not *have* to be random, but that's a safe default, and it allows the application to be stateless. (There's still the problem of reordered or missing messages, but I believe this should be out of scope.) I believe we should randomise.

---

I'm not sure how to handle the message. If we can confirm the following is legit, we should be good.

    uint8_t plain_text[] = "Lorem ipsum dolor sit amet, consectetur";

Hopefully the sign of `char*` won't matter. Otherwise I'm not sure how to fill a text we can actually wipe (string constants are supposed to be immutable, modifying them is undefined). In any case, if it works on our typical UNIX machines, I say we keep it as is. My prediction is that worst case, the user will get a warning. Plus, that text will necessarily come from some external source, so this declaration doesn't teach anything useful beyond the fact that `plain_text` is an input.

---

We should probably keep communication out. There are too many legitimate use cases for us to highlight any of them. I favour option 4. (I'm not really afraid of option 1, insofar as I intend to develop a full fledged file encryption utility at some point. We won't be able to avoid the back edge of this one.)

fscoto:
> One nitpick: perhaps use `.Xr arc4random_buf 3`? Might not be on the system, but if it is, it does have its manual section.

I *think* we've had this conversation before already, when doing the RNG section in intro(3monocypher) and decided against having `Xr` to pages that may not exist. If I recall correctly, that was because mandoc(1) with -Thtml would then try to generate links that are dead by definition. I can't seem to find the relevant GitHub issue anymore, or I'm just imagining things.

---

> I'm not sure how to handle the message. If we can confirm the following is legit, we should be good.
>
>     uint8_t plain_text[] = "Lorem ipsum dolor sit amet, consectetur";

I'd just like to point out that you're actually *doing* that in `chacha20_init_key`. In fact, I'm surprised that doing so works because [one might think that it wouldn't work on EBCDIC systems](https://stackoverflow.com/q/10938306). Either that or nobody's actually tried to get Monocypher running on an EBCDIC system yet.

Technically, casting to/from (unsigned) char\*/const char\* should always be legal; there might be some obscure C language gotchas though that I've missed when reading C11, § 6.5(7).

jsfierro:
@fscoto that section on arc4random_buf() looks good to me. It pushes the user to use a safe RNG and points to an explanation without being too wordy. Good call. 

I also think random nonces are the way to go here. It wards off a potential nonce reuse disaster and also indicates to the user that random nonces are safe to use in Monocypher, which is actually a really nice feature. Maybe a short comment explaining that it's safe with XChaCha20, but not with regular ChaCha20 is a good idea so that someone doesn't think  random nonces are _always_ safe if they use a different library. But I admit that this might be out of scope and too verbose here.

LoupVaillant:
Okay, looks good to me. Merging right away.

@fscoto: I forgot why you wouldn't use `.Xr`, sorry. I defer to you on this one.

@jsfierro the Chacha20 man page does explain pretty clearly why short nonces cannot be random. This explanation is absent from the `crypto_lock()` man page because its interface doesn't have short nonces. I presumed users would either use 24 random bytes, or a counter. Using, say, 8 random bytes and zeroing out the rest would be asking for trouble… Not sure we should add clutter about not doing *that*.

LoupVaillant:
@fscoto Okay, I forgot that you disabled the merge button. I can still force the merge, but I feel I need to ask: is this ready to merge?

fscoto:
I actually intended this to become a larger pull request with me going through every man page and liberally sprinkling randomness and other things everywhere, but I don't really care if you merge this now and I swap to a "one PR per page" approach instead.

LoupVaillant:
Ah, OK. Let's do this your way then: one PR with everything in it (do put use several commits if you want to).

fscoto:
I'd hit "ready for review" now, but you already did that a bit ago.

doc/man/man3/crypto_sign.3monocypher
```diff
@@ -150,32 +150,42 @@ return nothing.
...
 Sign a message:
 .Bd -literal -offset indent
-uint8_t       sk       [ 32]; /* Your secret key     */
-const uint8_t pk       [ 32]; /* Matching public key */
-const uint8_t message  [500]; /* Message to sign     */
-uint8_t       signature[ 64];
-crypto_sign(signature, sk, pk, message, 500);
+uint8_t       sk       [32]; /* Secret key from above          */
+const uint8_t pk       [32]; /* Matching public key            */
+const uint8_t message  [10] = "Lorem ipsu"; /* Message to sign */
```
LoupVaillant: Did you perhaps mean 11?

doc/man/man3/crypto_sign.3monocypher
```diff
@@ -150,32 +150,42 @@ return nothing.
...
 .Ed
 .Pp
 Check the above:
 .Bd -literal -offset indent
-const uint8_t pk       [ 32]; /* Their public key   */
-const uint8_t message  [500]; /* Signed message     */
-const uint8_t signature[ 64]; /* Signature to check */
-if (crypto_check(signature, pk, message, 500)) {
+const uint8_t pk       [32]; /* Their public key              */
+const uint8_t message  [10] = "Lorem ipsu"; /* Signed message */
```
LoupVaillant: There too

LoupVaillant:
I'd say good to merge. Just confirm that you wanted to exclude the final `\0` from the signature, and I can push the button.

fscoto:
Oh, no, that's just an off-by-one, see e202b00.

LoupVaillant:
Merged, thanks a lot!

PR #159, opened 2020-02-29
Document & test overlapping of key_exchange.

richwalm:
As discussed with Loup via e-mail, document and create a test for overlapping the shared_key & your_secret_key buffers for crypto_key_exchange().

Useful as in my application, there's no reason to keep around the secret key after obtaining a shared key, avoiding locking and wiping an extra buffer.

### DIFF
--- a/tests/test.c
+++ b/tests/test.c
@@ -9,7 +9,7 @@
 //
 // ------------------------------------------------------------------------
 //
-// Copyright (c) 2017-2019, Loup Vaillant
+// Copyright (c) 2017-2020, Loup Vaillant and Richard Walmsley
 // All rights reserved.
 //
 //
@@ -39,7 +39,7 @@
 //
 // ------------------------------------------------------------------------
 //
-// Written in 2017-2019 by Loup Vaillant
+// Written in 2017-2020 by Loup Vaillant and Richard Walmsley
 //
 // To the extent possible under law, the author(s) have dedicated all copyright
 // and related neighboring rights to this software to the public domain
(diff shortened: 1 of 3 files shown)


fscoto:
The documentation change seems good to me. As for the test code, *maybe* it would be worthwhile to have a `buf[64]` and loop from 0 to 32, offsetting one of the arguments by `i` to test partial overlapping.

(Suggestion for in-tree change: Add the same wording to `crypto_x25519` as well.)

I'd like to request that you add yourself with a copyright line and to the "Written by" line for the files you've changed, or that you make an explicit comment here for @LoupVaillant to do that after merging, explicitly acknowledging that you subject your work to BSD-2-clause and CC-0.

tests/test.c
```diff
@@ -660,6 +660,22 @@ static int p_argon2i_overlap()
...
+{
+    int status = 0;
+    FOR (i, 0, 5) {
+        u8 buf[32];
+        RANDOM_INPUT(sk, 32);
+        RANDOM_INPUT(pk, 32);
+        crypto_key_exchange(buf, sk, pk);
+        crypto_key_exchange(sk, sk, pk);
+        status |= memcmp(buf, sk, 32);
+    }
```
LoupVaillant: I agree with @fscoto on this one, ideally we should have a bigger buffer and test every possible overlap (62 tests total?). While partial overlap don't make sense, making sure they are possible makes sure the code resist some overly clever uses.

So:

    FOR (i, 0, 64) {
        u8 overlapping[96];
        u8 separate[32];
        RANDOM_INPUT(sk, 32);
        RANDOM_INPUT(pk, 32);
        memcmp(overlapping+32, sk, 32);
        crypto_key_exchange(separate, sk, pk);
        crypto_key_exchange(overlapping + i, overlapping + 32, pk);
        status |= memcmp(separate, overlapping, 32);
    }

(Not tested, not checked the consistency with the rest of the test suite, but you get the idea.)

richwalm:
Thanks. I've made those suggestions as mentioned.

LoupVaillant:
Perfect, merging right away, thanks!

PR #160, opened 2020-03-01
tests: fix typo in messages

fscoto:
Found through reading the patch to #159 and realizing that the typo was copied from surrounding code. See https://www.lexico.com/definition/overlap: “verb (overlaps, overlapping, overlapped)”

Completely trivial change, thus does not require changing file header.


LoupVaillant:
Good catch, thanks!

PR #162, opened 2020-03-24
Document Elligator and related plus X25519 inverse

fscoto:


doc/man/man3/crypto_curve_to_hidden.3monocypher
```diff
@@ -0,0 +1,283 @@
...
+which influences the final output of
+.Fn crypto_curve_to_hidden .
+.El
+.Pp
+The
+.Fa hidden ,
+.Fa curve ,
+and
+.Fa tweak
+arguments may overlap or point at the same buffer.
```
LoupVaillant: Not tweak, it's passed by value, and cannot possibly overlap with anything. I'd personally just say "The `hidden` and `curve` arguments are allowed to overlap".

doc/man/man3/crypto_curve_to_hidden.3monocypher
```diff
@@ -0,0 +1,283 @@
...
+.Sh RETURN VALUES
+.Fn crypto_curve_to_hidden
+returns 0 on success, -1 if the given
+.Fa curve
+argument is unsuitable for hiding.
```
LoupVaillant: Not sure we should document it, but in addition, this function has no effect if it fails. (I should test this).

doc/man/man3/crypto_curve_to_hidden.3monocypher
```diff
@@ -0,0 +1,283 @@
...
+return nothing; they cannot fail.
+.Sh EXAMPLES
+Generate a key pair manually using
+.Xr crypto_x25519_dangerous_small 3monocypher
+instead of its fast variant:
+.Bd -literal -offset indent
+uint8_t sk  [32]; /* Secret key output        */
+uint8_t pk  [32]; /* Hidden public key output */
+uint8_t tweak;    /* Random tweak input       */
+arc4random_buf(tweak, 1);
```
LoupVaillant: `&tweak`, I believe (tweak is not a pointer).

doc/man/man3/crypto_curve_to_hidden.3monocypher
```diff
@@ -0,0 +1,283 @@
...
+for (;;) {
+    arc4random_buf(sk, 32);
+    crypto_x25519_dangerous_small(pk, sk);
+    if (crypto_curve_to_hidden(pk, pk, tweak) == 0)
+        break;
+}
+/* Now save either:
+ * - the key pair (sk, pk), or
+ * - the secret key and tweak (sk, tweak) to reliably
+ *   re-generate the same value of pk.
```
LoupVaillant: I don't believe you should have any reason to anything except the secret key. You need to *send* the hidden public key (and record it in the transcript hash), but you should never need to send it twice, or even use it for any key exchange.

I believe this comment should just recommend we save the secret key, and send the public representative.

doc/man/man3/crypto_curve_to_hidden.3monocypher
```diff
@@ -0,0 +1,283 @@
...
+.Pp
+This function
+.Em helps
+build highly difficult-to-analyze protocols,
+but it is insufficient by itself:
+Other metadata, such as the amount of bytes sent in a packet or the size
+of the 32-byte random-looking string that represents the curve point
+itself, can be very strong indicators of the use of cryptography.
+Consider using appropriate padding algorithms, such as PADME,
+and obscure other metadata as much as possible.
```
LoupVaillant: :+1:

doc/man/man3/crypto_x25519_dangerous_fast.3monocypher
```diff
@@ -0,0 +1,135 @@
...
+.\" with this software.  If not, see
+.\" <https://creativecommons.org/publicdomain/zero/1.0/>
+.\"
+.Dd March 24, 2020
+.Dt CRYPTO_X25519_DANGEROUS_FAST 3monocypher
+.Os
+.Sh NAME
+.Nm crypto_x25519_dangerous_fast ,
+.Nm crypto_x25519_dangerous_small
+.Nd generation of Curve25519 points including low-order points
```
LoupVaillant: "_With_ a low order _component_".

We can't generate low order points with those because of clamping. The price for that is a slight, undetectable statistical bias. in the resulting key distribution (once you flip the sign at random).

Perhaps I wasn't clear, but the only danger with those functions is revealing 3 bits of the private scalar, and that doesn't give much leverage to anyone. (Hence my initial name, "dirty".

Did you misunderstand that bit, and maintain that we should spell those "dangerous?"

doc/man/man3/crypto_x25519_dangerous_fast.3monocypher
```diff
@@ -0,0 +1,135 @@
...
+These functions are used in public key generation for
+.Xr crypto_curve_to_hidden 3monocypher .
+.Sy This is a highly advanced feature ;
+unless you are reading this because you were referred here from
+.Xr crypto_curve_to_hidden 3monocypher ,
+.Sy you likely have no reason to be using these functions
+and are probably looking for
+.Xr crypto_key_exchange 3monocypher
+or
+.Xr crypto_x25519 3monocypher
```
LoupVaillant: crypto_key_exchange_public_key
    crypto_x25519_public_key

doc/man/man3/crypto_x25519_dangerous_fast.3monocypher
```diff
@@ -0,0 +1,135 @@
...
+but do the same with different code size and memory characteristics:
+.Fn crypto_x25519_dangerous_fast
+uses multiple large temporary variables and uses functions that are
+normally used internally for
+.Xr crypto_sign 3monocypher ;
+accordingly, it uses both more memory (for the temporary variables) and
+more code size (unless the signing code is already compiled in
+elsewhere).
+.Fn crypto_x25519_dangerous_small
+yields the same result, but does so using less time and memory at a
```
LoupVaillant: "less _code_ and memory"?

doc/man/man3/crypto_x25519_dangerous_fast.3monocypher
```diff
@@ -0,0 +1,135 @@
...
+large performance penalty compared to
+.Fn crypto_x25519_dangerous_fast .
+.Pp
+The resulting public keys
+.Em may
+be used with
+.Xr crypto_x25519 3monocypher
+and
+.Xr crypto_key_exchange 3monocypher ,
```
LoupVaillant: Actually, that's your only choice. They are *intended* to be used with those, by design.

doc/man/man3/crypto_x25519_dangerous_fast.3monocypher
```diff
@@ -0,0 +1,135 @@
...
+which clear the cofactor,
+but are dangerous for other uses because they contain low-order points,
+and may trip up other implementations of X25519.
```
LoupVaillant: Not sure what other use we could have.

Moreover, X25519 requires clearing the cofactor by clamping the scalar. And the fact that clamping does not preserve the prime order component means other implementations *will* clamp the scalar, clear the cofactor, and ignore the low order component.

Tripping up other implementations of X25519 is not possible. Not if they follow the paper, or the RFC. They *will* give different results with raw scalar multiplication, though.

doc/man/man3/crypto_x25519_dangerous_fast.3monocypher
```diff
@@ -0,0 +1,135 @@
...
+.Xr crypto_x25519 3monocypher ,
+.Xr intro 3monocypher
+.Sh HISTORY
+The
+.Fn crypto_x25519_dangerous_fast
+and
+.Fn crypto_x25519_dangerous_small
+functions first appeared in Monocypher 3.1.0.
+.Sh IMPLEMENTATION DETAILS
+The slow variant is approximately an entire 1.8 times slower than the
```
LoupVaillant: 2 times slower. 1.97 measured on my machine.

doc/man/man3/crypto_x25519_dangerous_fast.3monocypher
```diff
@@ -0,0 +1,135 @@
...
+fast variant.
+When considering that, on average, two calls to this function will be
+required for obtaining a valid keypair for
+.Xr crypto_curve_to_hidden 3monocypher ,
+this adds up to an
+.Em average
+effective slowdown for key pair generation of a factor of 3.6.
```
LoupVaillant: Neat trick: we can speed up *clean* key generation by using the fast "dangerous" version with a pre-clamped scalar (`scalar[0] &= 0xf8`). Not sure we want to document this, but users wanting to speed up ephemeral key pair generation can now use this.

Or we could add a dedicated function…

doc/man/man3/crypto_curve_to_hidden.3monocypher
```diff
@@ -0,0 +1,283 @@
...
+This is of interest for key exchange protocols that require
+indistinguishability from randomness, such as padded uniform random
+blobs (PURBs) and exponential blinding in oblivious pseudo-random
+functions (OPRFs).
+This is an
+.Em advanced feature
+\(en unless you are implementing an protocol that requires
+indistinguishability of all communications from random noise,
+consider
+.Xr crypto_key_exchange 3monocypher
```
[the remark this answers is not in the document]
fscoto: Already partially addressed in 3e5bd14.

doc/man/man3/crypto_x25519_dangerous_fast.3monocypher
fscoto: I'm reasonably sure that did not misunderstand that bit, but instead phrased it badly.

I do maintain that the names should include “dangerous” because it's a very loud alert that people are messing with things that are, in fact, probably not where they're intending to be, but could settle for “dirty” if accuracy matters more than the warning signal.

LoupVaillant: OK, I see. I believe I prefer "dirty", for a couple reasons:

- Your manual page already makes crystal clear how specialised and delicate those functions are. This is warning enough.
- "Dirty" is shorter than "dangerous" by 4 characters.
- The "danger" is minimal: worst case we are leaking 3 bits, that will never be used for anything else. Thus, "dirty" is more accurate.
- "Dangerous" can raise eyebrows, and get people to wonder why I considered it okay to put anything "dangerous" in a supposedly easy to use crypto library.

That last consideration is perhaps the most important one. I don't want a short sighted laser focus on those two function names to undermine the credibility of Monocypher. Something like:

    egrep "danger|warn|hack|TODO"

fscoto: Alright, okay, "dirty" it is after all.

LoupVaillant:
Looks good to merge. Anything else before I push the button?

fscoto:
I just pushed documentation for the new EdDSA->X25519 functions as well so that we can do everything in a single PR here.

doc/man/man3/crypto_from_eddsa_private.3monocypher
```diff
@@ -0,0 +1,132 @@
...
+These functions convert keys for use with
+.Xr crypto_sign 3monocypher
+(EdDSA with the BLAKE2b hash function)
+to keys for use with
+.Xr crypto_key_exchange 3monocypher
+and
+.Xr crypto_x25519 3monocypher .
+This may be useful in some resource-constrained contexts or when no
+other key is available (for example, when retrieving SSH public keys
+from GitHub and reusing the SSH private keys as X25519 public keys).
```
LoupVaillant: "reusing the SSH _public_ keys as X25519 public keys"

doc/man/man3/crypto_from_eddsa_private.3monocypher
```diff
@@ -0,0 +1,132 @@
...
+.Pp
+The
+.Fn crypto_from_eddsa_private
+function converts an EdDSA (with BLAKE2b) private key to an X25519
+private key.
+The
+.Fn crypto_from_eddsa_public
+function converts an EdDSA private key to an X25519 public key.
```
LoupVaillant: "converts an EdDSA _public_ key to an X25519 public key"

doc/man/man3/crypto_from_eddsa_private.3monocypher
```diff
@@ -0,0 +1,132 @@
...
+The arguments are:
+.Bl -tag -width Ds
+.It Fa eddsa
+The signing public key or private key to convert to a X25519 public key
+or private key, respectively.
+.It Fa x25519
+The converted private key or public key.
+.El
+.Pp
+The arguments may overlap or point at the same buffer.
```
LoupVaillant: (Note to self: add the test that makes sure they may.)

doc/man/man3/crypto_from_eddsa_private.3monocypher
```diff
@@ -0,0 +1,132 @@
...
+The
+.Fn crypto_from_eddsa_private
+and
+.Fn crypto_from_eddsa_public
+functions first appeared in Monocypher 3.1.0.
+.Sh SECURITY CONSIDERATIONS
+It is generally considered poor form to reuse the same key for different
+purposes.
+While this conversion is technically safe
+avoid these functions unless you are particularly resource-constrained
```
LoupVaillant: Maybe add a coma after "safe"?

doc/man/man3/crypto_from_eddsa_private.3monocypher
```diff
@@ -0,0 +1,132 @@
...
+or otherwise have a hard requirement nonetheless.
+It is otherwise an unnecessary risk factor.
```
LoupVaillant: Two "otherwise" in a row. I would remove the first one:

"or have a hard requirement nonetheless"

doc/man/man3/crypto_x25519_dirty_fast.3monocypher
```diff
@@ -0,0 +1,131 @@
...
+.Fn crypto_x25519_dirty_small
+yields the same result, but does so using less code and memory at a
+large performance penalty compared to
+.Fn crypto_x25519_dirty_fast .
+.Pp
+The resulting public keys are to be used with
+.Xr crypto_x25519 3monocypher
+and
+.Xr crypto_key_exchange 3monocypher ,
+which clear the cofactor,
```
LoupVaillant: Period instead of coma? (End of sentence.)

doc/man/man3/crypto_x25519_inverse.3monocypher
```diff
@@ -0,0 +1,102 @@
...
+.Pp
+This function is used, for example, with exponential blinding in
+oblivious pseudo-random functions (OPRFs).
+The arguments are:
+.Bl -tag -width Ds
+.It Fa blind_salt
+The output point.
+.It Fa private_key
+The private key (scalar) to use.
+This value's cofactor is cleared first,
```
LoupVaillant: Missed it the first time around: this value is *clamped*, according to the X25519 standard. That clears its cofactor all right, but it also affects the main factor. I would say:

    This value is clamped first,
    then inverted (modulo the curve order),
    then has its cofactor cleared,
    and finally is used for scalar multiplication.

(You may find a better way to say this, but you get the idea.)

doc/man/man3/intro.3monocypher
```diff
@@ -101,6 +101,11 @@ implements X25519, an elliptic curve Diffie Hellman key exchange
 algorithm based on Curve25519.
 X25519 derives a shared secret from two private/public key pairs.
 It is fast, simple, and relatively easy to implement securely.
+.Pp
+For specialized protocols that require indistinguishability from random
+noise,
+.Xr crypto_curve_to_hidden 3monocypher
+gives the option to disguise X25519 public keys as random noise.
```
LoupVaillant: _Ephemeral_ public keys to be more precise. Ratchet keys can be encrypted with the current session key, and static keys are reused and re-transmitted, and as such cannot possibly be indistinguishable from random —not the second time you send them over the network. Which is why static keys should also be encrypted if at all possible.

Not sure we nevertheless want to add the word "ephemeral". It would be more accurate, it would drop a keyword people can search for, but it could also confuse them…

fscoto: How about “one-time use (ephemeral)” instead?

LoupVaillant: Oh yes, _“one-time use (ephemeral)”_ is perfect.

LoupVaillant:
We're good now. Merging, any remaining problem can be fixed separately.

PR #171, opened 2020-04-18
Add extern C if C++ presents

stevefan1999-personal:
This fixes symbol compatibility if you have a project the compiles C and C++ separately

### DIFF
--- a/src/monocypher.h
+++ b/src/monocypher.h
@@ -57,6 +57,10 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 ////////////////////////
 /// Type definitions ///
 ////////////////////////
@@ -351,5 +355,8 @@ void crypto_x25519_inverse(uint8_t       blind_salt [32],
                            const uint8_t private_key[32],
                            const uint8_t curve_point[32]);
 
-
+#ifdef __cplusplus
+}
+#endif
+    
 #endif // MONOCYPHER_H


LoupVaillant:
Hi,

I was hesitating about adding this since basically the beginning of the project. The obvious advantage is that on a mixed project, with both C and C++ files, this would spare users the trouble of adding `extern C` themselves. But then I have two problems:

1. How standard is the __cplusplus pre-processor definition? While extremely widespread, Monocypher is very careful to be strictly conformant with C and C++. Anything that might be platform specific is avoided. To the point that to date, the only implementation defined behaviour in all Monocypher is the right shift of negative integers (avoiding that would have been very difficult).

2. Monocypher conforms to C *and C++*. There will be warnings about old style casts, but that's about it. If you compile Monocypher as C++, `extern C` no longer makes any sense. On the contrary, it's counter productive.

So yes, we could add `extern C`, and it would be more convenient for most uses in C++ projects. But it would establish Monocypher as C only, potentially reducing its portability (recall Microsoft compilers being basically C++ only). I'm not sure we want that.

Summoning the Chief Advisor Officer, @fscoto, and the Chief Testing Officer, @tankf33der (who tested Monocypher on numerous platforms): any thought?

fscoto:
As far as I'm aware, this *can* be worked around by covering the include directive that pulls in `monocypher.h` itself with `extern "C" {}`, but perhaps that's a bit too inconvenient.

---

> How standard is the __cplusplus pre-processor definition?

The `__cplusplus` pre-processor definition is mandated by the C++ standard; the C standard forbids setting it. C++98 final draft, §&nbsp;16.8(1):

> **16.8 Predefined macro names**
>
> 1 The following macro names shall be defined by the implementation:
>
> [...]
>
> `__cplusplus` The name `__cplusplus` is defined to the value `199711L` when compiling a C++ translation unit.

The C++98 standard also defines `extern "C"` linkage to exist in §&nbsp;7.5. This has not changed with the final draft for C++17 (N4659), §&nbsp;10.5.

C99 final draft (N1256), §&nbsp;6.10.8(5):

> **6.10.8 Predefined macro names**
>
> [...]
>
> 5 The implementation shall not predefine the macro `__cplusplus`, nor shall it define it in any standard header.

I therefore see no point, from a standards compliance perspective, not to go forward with this. I am unaware of any compiler that misbehaves with regards to this. Besides, I consider it extremely unlikely that a compiler will misbehave with `__cplusplus` because this kind of thing happens a lot, and especially historically happened a lot when C++ was still trying to get its own identity and mixed codebases were much more common.

---

> So yes, we could add extern C, and it would be more convenient for most uses in C++ projects. But it would establish Monocypher as C only, potentially reducing its portability (recall Microsoft compilers being basically C++ only). I'm not sure we want that.

It would definitely put Monocypher on a fairly slippery slope of disregarding C++ more than we already have. I would, however, be lying if I didn't admit that making this change over and over hasn't been annoying to me, either.

Generally, you will need to make *some* change to work around Monocypher as-is in a C++ project:

1. Add an `extern "C"` block around every `#include "monocypher.h"`.
2. Rename `monocypher.c` to `monocypher.cpp` and possibly `monocypher.h` to `monocypher.hpp` to compile it as native C++. This is, obviously, not very convenient if Monocypher is compiled as an OS library, and downright foolish to attempt if installed from a package.

This requirement of a change is somewhat suboptimal because (a) it is inconvenient, (b) people may—and probably should—have inhibitions to changing cryptography-related code, no matter how minor the change, and (c) method&nbsp;2 above needs to be repeated for every update of Monocypher.

The README still holds that Monocypher compiles as C++, so we *are* still making a promise that it is supposed to work.

Ignoring all my thinly-veiled rationalizations: I've run into this in the past, and it's annoyed me, but I didn't want to be the one to bring it up because maybe it's just me who is annoyed.

As things stand, I'm **in favor**, but willing to scale back if more arguments are made as to why it's a bad idea.

LoupVaillant:
OK, I'm sold. Merging.

@stevefan1999-personal thanks for the patch.

PR #173, opened 2020-04-22
Make tarball generation reproducible

fscoto:
There's a *lot* of other environmental damage that can't really be avoided, so reprodicibility depends on the exact setup used to generate the tarball anyway:

- at least OpenBSD sed(1) apparently munges the README *slightly* differently
- mandoc(1) versions *must not* differ or the HTML generated by doc2html will differ
- the operating system reported in mandoc(1) *must not* differ or the HTML generated by doc2html will differ
- probably more that I didn't catch after giving up at cross-OS reproducibility

So for anyone to be able to actually reproduce a release tarball, they need to know the exact setup. Perhaps building on a specific public Docker image and specifying this might be the easiest way to fixate and publicize this.

### DIFF
--- a/dist.sh
+++ b/dist.sh
@@ -12,7 +12,7 @@
 # ------------------------------------------------------------------------
 #
 # Copyright (c) 2019, Loup Vaillant
-# Copyright (c) 2019, Fabio Scotoni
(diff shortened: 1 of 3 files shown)


LoupVaillant:
Looks good to me. Merging, thanks.

PR #178, opened 2020-05-24
doc: Fix .In for optional code

fscoto:


LoupVaillant:
Good catch, thanks!

fscoto:
I'd be considerably happier if I did such good work that I didn't *need* to catch these things every time I try to use the documentation I wrote myself, but alas, the universe disagrees.

Issue #189, opened 2020-09-23
Documentation: Confusing statement on signature malleability


snej:
The [documentation on signatures](https://monocypher.org/manual/sign#Signature_malleability) says:

>EdDSA signatures are not unique like cryptographic hashes. For any given public key and message, there are many possible valid signatures. Some of them require knowledge of the private key. **Others only require knowledge of an existing signature.** Observing a valid signature only proves that someone with knowledge of the private key signed the message at some point. Do not rely on any other security property.

I found the boldfaced phrase confusing. At first it sounded like a signature could be modified to act as a signature of a different message, but then I realized that would be a fatal bug. I think what you mean is that, given a valid signature of message X, it's possible for someone without knowledge of the private key to _alter some byte(s) of the signature_ so that it's still a valid signature of X. Is that accurate?

I think I would rephrase this warning as something like:

>EdDSA signatures are not unique like cryptographic hashes. For any given public key and message, there are many possible valid signatures. **A signature can be modified by someone without knowledge of the private key, while remaining valid. It's therefore not valid to assume that, if two signatures are byte-for-byte different, that they must sign two different messages.** Observing a valid signature only proves that someone with knowledge of the private key signed the message at some point. Do not rely on any other security property.

This also begs the question of whether two calls to `crypto_sign` with the same message and key-pair always produce the same output bytes. I assume they do, but it's not stated. (libSodium's docs do make this guarantee.)

fscoto:
> I think what you mean is that, given a valid signature of message X, it's possible for someone without knowledge of the private key to alter some byte(s) of the signature so that it's still a valid signature of X. Is that accurate?

As far as I know, that is indeed what this is supposed to mean. To the best of my knowledge, this part is about [signature malleability](https://tools.ietf.org/html/rfc8032#section-8.4), the details of which are fairly boring and involve :math:. Monocypher has had a check against malleability since 24f4be7acc3ec7ff613715a7a97597e587f6d6d8, so in that sense the passage is irrelevant; however, I believe it was kept as a warning about EdDSA in general because Monocypher may be used together in a larger systems with other libraries working with EdDSA. @LoupVaillant, please correct me if I remember wrong.

[...]

PR #191, opened 2020-09-26
HISTORY: note 2.0.5 rejecting modified signatures

fscoto:
Did some dumpster diving to figure this out. In retrospect, I'm surprised there wasn't a major version spike after this, but it's not like signatures generated by abusing signature malleability ever happen with legitimate signers in the first place.

There doesn't seem to be a good term to refer to a signature that is valid but has a greater *S* value than *L* without dragging the whole explanation of EdDSA in.

---

Change introduced in 974e55d21c1fac7a2e21f91cb7174601b653180a and
24f4be7acc3ec7ff613715a7a97597e587f6d6d8.

The actual reasons to introduce this were actually performance-related.

Sparked by #189.


LoupVaillant:
That's perfect, thanks.

PR #195, opened 2020-10-04
doc: clarify signature malleability in crypto_sign

### LINKED ISSUE #189: Documentation: Confusing statement on signature malleability

fscoto:
Fixes #189; see the discussion there for details.

(Apologies for the delay, I forgot for a bit that the onus to PR was on me and just kept wondering why Loup wasn't committing anything.)

### DIFF
--- a/doc/man/man3/crypto_sign.3monocypher
+++ b/doc/man/man3/crypto_sign.3monocypher
@@ -220,14 +220,23 @@ Monocypher 0.3;
 it was fixed in Monocypher 1.1.1 and 2.0.4.
 .Sh SECURITY CONSIDERATIONS
 .Ss Signature malleability
-EdDSA signatures are not unique like cryptographic hashes.
-For any given public key and message, there are many possible valid
-signatures.
-Some of them require knowledge of the private key.
-Others only require knowledge of an existing signature.
-Observing a valid signature only proves that someone with knowledge of
-the private key signed the message at some point.
-Do not rely on any other security property.
+Signature malleability is the ability of an attacker to produce a valid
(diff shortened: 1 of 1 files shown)


LoupVaillant:
Perfect, thank you. Don't worry about the delay, there was really no rush.

PR #203, opened 2020-12-27
crypto_ietf_chacha20: note nonce overflow handling

fscoto:
IETF ChaCha20 has a 32-bit counter. This means a practical limit of 256 GiB of data for each nonce. Additionally, IETF QUIC seems to require being able to handle 0xffffffff (I-D.draft-ietf-quic-tls-33 §&nbsp;5.4.4), thus getting very close to the overflow, though not triggering it.

Unlike libsodium and other libraries, we do not have the option to panic and take down whatever process is running the code and triggering the overflow condition because Monocypher is neither allowed to use the C standard library nor allowed to invoke undefined behavior to cause a crash; the applicable RFC provides no guidance what to do in this case, either.

Therefore, staying within the (nonce, counter) limits is necessarily application responsibility; it is an invariant that, when voided, Monocypher is allowed to do anything, similar to the non-guarantee we make for the `crypto_blake2b` family and the `crypto_argon2i` family.

While already here, fix the wrong function prototype in the synopsis.

### DIFF
--- a/doc/man/man3/advanced/crypto_ietf_chacha20.3monocypher
+++ b/doc/man/man3/advanced/crypto_ietf_chacha20.3monocypher
@@ -65,7 +65,7 @@
 .Fa "const uint8_t key[32]"
 .Fa "const uint8_t nonce[12]"
 .Fc
-.Ft void
+.Ft uint32_t
 .Fo crypto_ietf_chacha20_ctr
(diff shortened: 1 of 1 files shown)


LoupVaillant:
Perfect, thank you very much. Merging right away.

PR #204, opened 2021-02-05
Add API exports

xoofx:
Hello there!

This PR is adding API exports, so that symbols are correctly exported from the shared lib (specially on Windows+MSVC)

I put the `MONOCYPHER_EXPORT_API` on the previous line of function declarations to avoid noisy changes on the API, let me know if you want it on the same line.

### DIFF
--- a/src/monocypher.h
+++ b/src/monocypher.h
@@ -57,6 +57,12 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#ifdef WIN32
+#define MONOCYPHER_EXPORT_API __declspec(dllexport)
+#else
+#define MONOCYPHER_EXPORT_API __attribute__((visibility("default")))
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -119,24 +125,30 @@ typedef crypto_sign_ctx crypto_check_ctx;
 // -------------------------
 
 // Return 0 if a and b are equal, -1 otherwise
+MONOCYPHER_EXPORT_API
 int crypto_verify16(const uint8_t a[16], const uint8_t b[16]);
+MONOCYPHER_EXPORT_API
 int crypto_verify32(const uint8_t a[32], const uint8_t b[32]);
+MONOCYPHER_EXPORT_API
 int crypto_verify64(const uint8_t a[64], const uint8_t b[64]);
 
 // Erase sensitive data
 // --------------------
 
 // Please erase all copies
+MONOCYPHER_EXPORT_API
 void crypto_wipe(void *secret, size_t size);
 
 
 // Authenticated encryption
 // ------------------------
+MONOCYPHER_EXPORT_API 
 void crypto_lock(uint8_t        mac[16],
                  uint8_t       *cipher_text,
                  const uint8_t  key[32],
                  const uint8_t  nonce[24],
                  const uint8_t *plain_text, size_t text_size);
+MONOCYPHER_EXPORT_API
 int crypto_unlock(uint8_t       *plain_text,
                   const uint8_t  key[32],
                   const uint8_t  nonce[24],
@@ -144,12 +156,14 @@ int crypto_unlock(uint8_t       *plain_text,
                   const uint8_t *cipher_text, size_t text_size);
 
 // With additional data
+MONOCYPHER_EXPORT_API
 void crypto_lock_aead(uint8_t        mac[16],
                       uint8_t       *cipher_text,
                       const uint8_t  key[32],
                       const uint8_t  nonce[24],
                       const uint8_t *ad        , size_t ad_size,
                       const uint8_t *plain_text, size_t text_size);
+MONOCYPHER_EXPORT_API
 int crypto_unlock_aead(uint8_t       *plain_text,
                        const uint8_t  key[32],
                        const uint8_t  nonce[24],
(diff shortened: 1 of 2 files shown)


xoofx:
Side question: Would you accept a separate PR to add support for building with cmake? (only shared libs, not tests). The cmake fiels are very small, just a couple of lines...
I'm using it to cross-build between windows, mac and linux.

LoupVaillant:
Ah, yes, _Windows_. I'm torn about this, I'd like to have @fscoto's opinion.

Monocypher was initially designed to be included in the source code of the projects that use it, so there's no need for such platform specific shenanigans. You are addressing a very specific scenario: exporting the symbols of a low-level cryptographic library in a Windows DLL. I'm no sure we want to shoulder this specific use case ourselves. Users who really need it can always add the `__declspec` declarations themselves as they need it. Tedious perhaps, but still easy and safe.

One very important thing to note is that not all Monocypher users on Windows will want to export its symbols in the first place. I can give an example at work, where we use Monocypher in some core library, but its symbols remain hidden from the outside. We don't want to give users of our library access to Monocypher. We want to give them access to the encrypted channel.

If we go through with this, we need to clarify a couple points:

* I think defining an attribute on every platform besides WIN32 will hurt portability. Not everyone will be using GCC or LLVM, especially in embedded targets, which are our most important users so far. We should probably cause the `#else` clause to simply neutralise `MONOCYPHER_EXPORT_API` with `#define MONOCYPHER_EXPORT_API`.

* You handled export, but I believe there's import to think about as well. The way I've seen it done at work, was that we import by default (so users including `monocypher.h` can tell their compiler to import), and we include in the source file a definition that causes us to export instead. I'm not sure however how possible without involving project-specific files.

---

Re: CMake: for the longest time, I resisted including any build system but the bare minimum. If I include CMake, what's stopping us from including the autotools, Buck (which i have refused in the past), or any other build system? The idea is that we have sufficiently few files (1 compilation unit, 2 if you add the optional parts), that it's easy for users to just include them in their project as if they wrote them. So easy in fact that I believe that putting Monocypher where it can be found, and adding the suitable `findpackage` declaration in the project that use is not even easier.

Sure, since CMake would be so short (at least if we don't include the test suite & the benchmarks), it's very tempting to just add it. But then, if it is so short, can't end users easily add it themselves? Again, Monocypher is not Libsodium. It's so easy to build that it's not clear adding support for any given build system really saves work.

Again, I'd like to hear @fscoto's thoughts on this.

xoofx:
> I think defining an attribute on every platform besides WIN32 will hurt portability. Not everyone will be using GCC or LLVM, especially in embedded targets, which are our most important users so far. We should probably cause the #else clause to simply neutralise MONOCYPHER_EXPORT_API with #define MONOCYPHER_EXPORT_API.

Fair enough. Do you want to hide the entire export through a define then instead? (e.g `MONOCYPHER_EXPORT_API_ENABLE`)

> If I include CMake, what's stopping us from including the autotools,

No worries, I will handle CMakefiles in my repo.

xoofx:
Just to add the context of my usage: planning to use Monocypher from a microcontroller (e.g ESP8266) to connect to a Web service developed in C#

xoofx:
Thinking about it, you are right, I can handle the wrapping on my side without touching Monocypher and that will allow also to simplify the API for binding to a language like C#, so it's actually more suitable. 
I'm closing this issue.

fscoto:
Inevitably, there comes a point in any sufficiently large where the dllexports are added with a macro hack. Clearly, Monocypher's time has come.

I'd normally first try to argue that there may be a case where you don't want to export, but @LoupVaillant's already done that for me. So allow me to instead point out that you don't *have* to litter any code with attributes just to handle Windows: [DEF files allow picking exports selectively and are out-of-source-file](https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-def-files?view=msvc-160).

Another thing specific to this, agreeing with Loup there: The fallback `__attribute__((visibility("default")))` introduces attributes. Attributes are non-portable by design because the C and C++ standards don't contain them in the first place (unlike bracket-delimited attributes, which have been in C++ for a while and seem to be making their way into C2x). That would go immediately counter to Monocypher's stated goal of portability.

About CMake/Buck/..., I'm honestly on the fence. We *do* have something that is kind of build-system-specific because we do create a pc file for pkg-config, which is primarily also to aid people packaging it in Linux distributions to get a mostly uniform result there. But pkg-config has poor integration with e.g. Windows. Ultimately, I think that may require revisiting. CMake seems to have a FindPkgConfig module though, so at least on non-Windows, that may be a solution for CMake integration.

xoofx:
> I'd normally first try to argue that there may be a case where you don't want to export, but @LoupVaillant's already done that for me. So allow me to instead point out that you don't have to litter any code with attributes just to handle Windows: DEF files allow picking exports selectively and are out-of-source-file.

Yeah, I'm aware of that, but it doesn't come in sync with the source code. You need to update your file on your side, know which function was added...etc.
In practice, you would end-up parsing the header file and create a def file from it.

> Another thing specific to this, agreeing with Loup there: T

Yes, indeed, I only thought about my usage obviously 😅 

> About CMake/Buck/..., I'm honestly on the fence.

That's ok. I can handle everything outside of Monocypher repo, so my PR was unnecessary. Sorry for the noise, and thanks for your answers.

LoupVaillant:
@fscoto, thank you for your input. @xoofx, don't worry about the noise, you were raising legitimate issues. It's also very nice to see I have users who care enough to try and make the project better.

> Yeah, I'm aware of that, but it doesn't come in sync with the source code. You need to update your file on your side, know which function was added...etc.

While it may not be obvious from the version number, Monocypher's API is very stable. Even the _ABI_ is quite stable. Things were added over the years, but very little _changed_. And as a bonus, every major branch is on long term support (I believe I promised something like "until 2024"). Moreover, I'm pretty sure now that Monocypher support all the low level primitives I need. Higher level stuff like authenticated key exchange, PAKE, safe streaming interfaces for encryption, network protocols, will be tackled in separate projects.

> On Windows/desktop linux/mac, I would load Monocypher native shared library from C#.

That is *extremely* interesting. If you intend to publish your C# bindings, they would be a very nice addition to our [current list](https://monocypher.org/download/).

---

About Cmake, Buck, etc. I will record here that the issue is not set in stone. We will consider adding direct support, for the reasons @fscoto has mentioned. If we keep them to a minimum (assume we're dealing with a tarball, compile the library and the basic tests only), we may be looking at maybe one little file per build system, which might actually be manageable.

I originally didn't want to fall prey to scope creep, but a more rigorous cost/benefit analysis won't hurt. Thank you @xoofx  for raising this issue to our attention. That definitely wasn't noise.

PR #213, opened 2021-06-11
Canonicalize spellings of algorithms

fscoto:
Inspired by [this tweet by Jean-Philippe Aumasson](https://twitter.com/veorq/status/1396728032883322884). I believe it makes a positive impression to actually spell names correctly.

I didn't touch the code mainly for reasons of not intruding upon @LoupVaillant's territory, but it may be reasonable to also adjust comments and text case output as well.


LoupVaillant:
Oops. Should have know about BLAKE2. I had no idea about ChaCha. Time to update my spell checkers I guess.
In any case, thanks for the correction. Merging right away.

PR #220, opened 2021-09-08
doc: Argon2 now has an RFC, not just a draft

fscoto:


LoupVaillant:
Thanks for watching out. I’ll update the web site soon.

PR #232, opened 2022-02-12
doc: Mention hashing the public keys

### LINKED ISSUE #230: Using the shared secret directly
Hi @LoupVaillant,

In reference to Curve25519 and Curve448, [RFC 7748](https://www.rfc-editor.org/rfc/rfc7748#section-7) says:

> Designers using these curves should be aware that for each public key, there are several publicly computable public keys that are equivalent to it, i.e., they produce the same shared secrets.  Thus using a public key as an identifier and knowledge of a shared secret as proof of ownership (without including the public keys in the key derivation) might lead to subtle vulnerabilities.

Similarly, on the [Point*scalar multiplication](https://doc.libsodium.org/advanced/scalar_multiplication#usage) page of the libsodium documentation, Frank notes the following:

> q represents the X coordinate of a point on the curve. As a result, the number of possible keys is limited to the group size (≈2^252), which is smaller than the key space.
> 
> For this reason, and to mitigate subtle attacks due to the fact many (p, n) pairs produce the same result, using the output of the multiplication q directly as a shared key is not recommended.
> 
> A better way to compute a shared key is h(q ‖ pk1 ‖ pk2), with pk1 and pk2 being the public keys.
> 
> By doing so, each party can prove what exact public key they intended to perform a key exchange with (for a given public key, 11 other public keys producing the same shared secret can be trivially computed).

This suggests it would be safest to have `crypto_key_exchange()` use BLAKE2b instead of HChaCha20 and include both public keys in the calculation of the shared key, a bit like the `crypto_kx()` API in libsodium. At the very least, I think it should be noted in the documentation. I will be mentioning it in my guidelines document as it's definitely something people overlook. It's something I need to change in Kryptor for v4 as well.

samuel-lucas6:
Closes #230. May need some rewording.

### DIFF
--- a/doc/man/man3/crypto_key_exchange.3monocypher
+++ b/doc/man/man3/crypto_key_exchange.3monocypher
@@ -12,6 +12,7 @@
 .\" Copyright (c) 2017-2018 Michael Savage
 .\" Copyright (c) 2017, 2019-2021 Fabio Scotoni
 .\" Copyright (c) 2020 Richard Walmsley
+.\" Copyright (c) 2022 Samuel Lucas
 .\" All rights reserved.
 .\"
 .\"
@@ -41,8 +42,8 @@
 .\"
 .\" ----------------------------------------------------------------------------
 .\"
-.\" Written in 2017-2021 by Loup Vaillant, Michael Savage, Fabio Scotoni and
-.\" Richard Walmsley
+.\" Written in 2017-2022 by Loup Vaillant, Michael Savage, Fabio Scotoni,
+.\" Richard Walmsley and Samuel Lucas
 .\"
 .\" To the extent possible under law, the author(s) have dedicated all copyright
 .\" and related neighboring rights to this software to the public domain
@@ -179,6 +180,14 @@ If either of the long-term secret keys leaks, it may compromise
 .Em all past messages .
 This can be avoided by using protocols that provide forward secrecy,
 such as the X3DH key agreement protocol.
+.Pp
+Many (private, public) key pairs produce the same shared secret.
+Therefore, not including the public keys in the key derivation can 
+lead to subtle vulnerabilities. This can be avoided by hashing the
+shared secret concatenated with both public keys. For example,
+.Fn BLAKE2b(shared_secret || your_pk || their_pk)
+using
+.Xr crypto_blake2b 3monocypher .
 .Sh IMPLEMENTATION DETAILS
 .Fn crypto_key_exchange_public_key
 is an alias to
--- a/doc/man/man3/crypto_key_exchange.3monocypher
+++ b/doc/man/man3/crypto_key_exchange.3monocypher
@@ -53,7 +53,7 @@
 .\" with this software.  If not, see
 .\" <https://creativecommons.org/publicdomain/zero/1.0/>
 .\"
-.Dd June 11, 2021
+.Dd February 12, 2022
 .Dt CRYPTO_KEY_EXCHANGE 3MONOCYPHER
 .Os
 .Sh NAME
@@ -182,10 +182,12 @@ This can be avoided by using protocols that provide forward secrecy,
 such as the X3DH key agreement protocol.
 .Pp
 Many (private, public) key pairs produce the same shared secret.
-Therefore, not including the public keys in the key derivation can 
-lead to subtle vulnerabilities. This can be avoided by hashing the
-shared secret concatenated with both public keys. For example,
-.Fn BLAKE2b(shared_secret || your_pk || their_pk)
+Therefore, not including the public keys in the key derivation can
+lead to subtle vulnerabilities.
+This can be avoided by hashing the shared secret concatenated with
+both public keys.
+For example,
+.D1 BLAKE2b(shared_secret || your_pk || their_pk)
 using
 .Xr crypto_blake2b 3monocypher .
 .Sh IMPLEMENTATION DETAILS


doc/man/man3/crypto_key_exchange.3monocypher
```diff
@@ -179,6 +180,14 @@ If either of the long-term secret keys leaks, it may compromise
 .Em all past messages .
 This can be avoided by using protocols that provide forward secrecy,
 such as the X3DH key agreement protocol.
+.Pp
+Many (private, public) key pairs produce the same shared secret.
+Therefore, not including the public keys in the key derivation can 
+lead to subtle vulnerabilities. This can be avoided by hashing the
+shared secret concatenated with both public keys. For example,
```
fscoto, line 187: troff is sensitive to periods.

```
This is a sentence.
This is another sentence.
```

[will use sentence spacing](https://mandoc.bsd.lv/mdoc/intro/punctuation.html), but

```
This is a sentence. This is another sentence.
```

will not.

Please add newlines after each period.

doc/man/man3/crypto_key_exchange.3monocypher
```diff
@@ -41,8 +42,8 @@
 .\"
 .\" ----------------------------------------------------------------------------
 .\"
-.\" Written in 2017-2021 by Loup Vaillant, Michael Savage, Fabio Scotoni and
-.\" Richard Walmsley
+.\" Written in 2017-2022 by Loup Vaillant, Michael Savage, Fabio Scotoni,
+.\" Richard Walmsley and Samuel Lucas
```
fscoto, line 46: Please update `.Dd` on line 56, too, since the manual contents changed.

doc/man/man3/crypto_key_exchange.3monocypher
```diff
@@ -179,6 +180,14 @@ If either of the long-term secret keys leaks, it may compromise
 .Em all past messages .
 This can be avoided by using protocols that provide forward secrecy,
 such as the X3DH key agreement protocol.
+.Pp
+Many (private, public) key pairs produce the same shared secret.
+Therefore, not including the public keys in the key derivation can 
+lead to subtle vulnerabilities. This can be avoided by hashing the
+shared secret concatenated with both public keys. For example,
+.Fn BLAKE2b(shared_secret || your_pk || their_pk)
```
fscoto, line 188: [`.Fn`](https://man.openbsd.org/mdoc.7#Fn) is used only for function names. `BLAKE2b()` is not a function.

Instead use [`.Dl`](https://man.openbsd.org/mdoc.7#Dl) (indented) or [`.Ql`](https://man.openbsd.org/mdoc.7#Ql).

fscoto (changes requested):
Should we also mark the function as deprecated and likely to be removed in a future release considering it's just... not good?

samuel-lucas6:
Sorry, not used this before. Thanks @fscoto :)

samuel-lucas6:
This unfortunately doesn't seem to be formatted properly on the website:

temp

LoupVaillant:
@samuel-lucas6 good catch, thanks. Both issues should be fixed now.

samuel-lucas6:
> @samuel-lucas6 good catch, thanks. Both issues should be fixed now.

I'm afraid the 'Security Considerations' fix doesn't seem to have worked. I've also noticed the ChaCha20, IETF ChaCha20, and HChaCha20 pages in 'Advanced' and 'Deprecated' haven't got the second 'c' capitalised, which was changed in the text of these pages.

LoupVaillant:
> I'm afraid the 'Security Considerations' fix doesn't seem to have worked.

Refresh with Ctrl+F5, or hold the control key when clicking on the "refresh" button. I have noticed that you often need such deep refreshes to make the fix visible (it’s a CSS fix, not an HTML one).

> I've also noticed the ChaCha20, IETF ChaCha20, and HChaCha20 pages in 'Advanced' and 'Deprecated' haven't got the second 'c' capitalised

Oops, I forgot to update the side bar… Fixed now.

samuel-lucas6:
> Refresh with Ctrl+F5, or hold the control key when clicking on the "refresh" button. I have noticed that you often need such deep refreshes to make the fix visible (it’s a CSS fix, not an HTML one).

Tried that on three browsers, and it sadly didn't work.

One more thing (sorry) is that the 'Home' page link to 'Public Key Cryptography' no longer works.

LoupVaillant:
> the 'Home' page link to 'Public Key Cryptography' no longer works.

Oh, yes, shuffling the pages around broke the link. Correcting now, thanks.

> Tried that on three browsers, and it sadly didn't work.

OK, now I'm very, *very* surprised. Here's what I'm seeing right now:

Screenshot from 2022-02-16 00-13-44

Aren't seeing this result? If not, can you take a look at https://monocypher.org/css/style.css and confirm the presence of the following code?

```CSS
div.Bd {
    margin: -1.5rem 0 0 1.5rem;
}
```

Quite hack-ish, but I think that's what most closely imitates the layout on the man page itself (newline, indent, no vertical spacing). I could change the font to fixed size, but hesitated to do so because this is not valid C code.

samuel-lucas6:
> Aren't seeing this result? If not, can you take a look at https://monocypher.org/css/style.css and confirm the presence of the following code?

Ok that is what I'm seeing, and the CSS is present. I was expecting the last bit to be on one line.

PR #233, opened 2022-02-13
Deprecate crypto_key_exchange

fscoto:
Continuation of #230 and #232.

Please also look over the commit message so I can reword it and force push if necessary.


doc/man/man3/advanced/crypto_hchacha20.3monocypher
```diff
@@ -114,7 +114,6 @@ crypto_wipe(in , 16);
 .Ed
 .Sh SEE ALSO
 .Xr crypto_chacha20_encrypt 3monocypher ,
```
LoupVaillant, line 116: This one should be replaced by `crypto_chacha20` in a separate commit. (`crypto_chacha20_encrypt` is deprecated, and should not be mentioned here).

LoupVaillant (approved):
Good to merge as far as I am concerned, including the message commit.

Unrelated to this, I just noticed that the manual failed to properly deprecate `crypto_chacha20_encrypt`, we should make another pass in a separate commit.

PR #235, opened 2022-03-03
doc/crypto_hchacha20: Fix typo.

samuel-lucas6:
Please let me know if updating the copyright is unnecessary in this case.

### DIFF
--- a/doc/man/man3/advanced/crypto_hchacha20.3monocypher
+++ b/doc/man/man3/advanced/crypto_hchacha20.3monocypher
@@ -11,6 +11,7 @@
 .\" Copyright (c) 2017-2019 Loup Vaillant
 .\" Copyright (c) 2017-2018 Michael Savage
 .\" Copyright (c) 2019-2022 Fabio Scotoni
+.\" Copyright (c) 2022 Samuel Lucas
 .\" All rights reserved.
 .\"
 .\"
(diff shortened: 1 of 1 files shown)


doc/man/man3/advanced/crypto_hchacha20.3monocypher
```diff
@@ -105,7 +107,7 @@ Simple hash:
 .Bd -literal -offset indent
 uint8_t key[32]; /* Must have enough entropy           */
 uint8_t in [16]; /* Does not have to be random         */
-uint8_t out[32]; /* Will be random iff the above holds */
+uint8_t out[32]; /* Will be random if the above holds */
```
cfogelklou: IFF is a short version of "If and only if" so the original might not be a typo.

cfogelklou (changes requested):
Restore your "typo fix" to "iff". The original author has used iff elsewhere in the code, so he/she means "if and only iff" and it is not a typo.
Also, this is your only contribution so probably this whole PR is unnecessary :).

samuel-lucas6:
Thanks for pointing that out. I proofread the entire documentation a while back but Loup did all the actual edits, so it's possible I reported this being wrong but nothing was changed. I've not seen 'iff' used before I don't think.

cfogelklou:
Hi Samuel, I am not an official contributor or anything - Perhaps you should leave your PR open in case Loup decides it actually was a typo :). Best, CF.

LoupVaillant:
Hi @samuel-lucas6, Thanks for the attempt.

I believe the "iff" was intentional, and did meant "if and only if". In this particular case, it’s really an equivalence: the buffer will be random if we have enough entropy in the seed, and will not be random if we don’t have enough entropy. I personally vote for keeping the "iff".

PR #237, opened 2022-03-05
Add Github Actions CI, following the instructions from the readme.md to run basic tests. Add CI badge for Github Actions to readme.md.

cfogelklou:
Since Github Actions on Ubuntu is free for Open Source, why not use it? :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     name: Build monocypher and run unit tests
     strategy:
       matrix:
-        platform: [ubuntu-latest]
+        platform: [ubuntu-latest] #macos-latest could also be here, but not free for Open Source
     runs-on: ${{ matrix.platform }}
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,3 +1,54 @@
+# This file is dual-licensed.  Choose whichever licence you want from
+# the two licences listed below.
+#
+# The first licence is a regular 2-clause BSD licence.  The second licence
+# is the CC-0 from Creative Commons. It is intended to release Monocypher
+# to the public domain.  The BSD licence serves as a fallback option.
+#
+# SPDX-License-Identifier: BSD-2-Clause OR CC0-1.0
+#
+# ------------------------------------------------------------------------
+#
+# Copyright (c) 2017-2022, Loup Vaillant
+# All rights reserved.
+#
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the
+#    distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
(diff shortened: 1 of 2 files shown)


cfogelklou:
If you merge this, subsequent PRs and merges to master will be run via CI and a badge will be placed in the readme.md indicating the CI status.
Github Actions provides free CI using Ubuntu for open source projects (which this is)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Run speed tests
+        run: |
+          make clean && make speed
+
+      - name: Run code coverage tests
+        run: |
+          ./tests/coverage.sh
+
+      # This test fails with a crash. The set +e handles this for now, but the source of the crash
+      # should be found and fixed.
```
LoupVaillant, line 63: Note that those tests work on my machine. On Github however, it would seem Valgrind does not work as it should:

```
vex amd64->IR: unhandled instruction bytes: 0x62 0xF1 0xFD 0x28 0x6F 0x35 0x9 0x5E 0x13 0x0
vex amd64->IR:   REX=0 REX.W=0 REX.R=0 REX.X=0 REX.B=0
vex amd64->IR:   VEX=0 VEX.L=0 VEX.nVVVV=0x0 ESC=NONE
vex amd64->IR:   PFX.66=0 PFX.F2=0 PFX.F3=0
==3862== valgrind: Unrecognised instruction at address 0x15d28d.
==3862==    at 0x15D28D: vector_of_string (in /home/runner/work/Monocypher/Monocypher/test.out)
==3862==    by 0x15D764: next_input (in /home/runner/work/Monocypher/Monocypher/test.out)
==3862==    by 0x15B839: chacha20 (in /home/runner/work/Monocypher/Monocypher/test.out)
==3862==    by 0x15D8B0: vector_test (in /home/runner/work/Monocypher/Monocypher/test.out)
==3862==    by 0x159279: main (in /home/runner/work/Monocypher/Monocypher/test.out)
==3862== Your program just tried to execute an instruction that Valgrind
==3862== did not recognise.  There are two possible reasons for this.
==3862== 1. Your program has a bug and erroneously jumped to a non-code
==3862==    location.  If you are running Memcheck and you just saw a
==3862==    warning about a bad jump, it's probably your program's fault.
==3862== 2. The instruction is legitimate but Valgrind doesn't handle it,
==3862==    i.e. it's Valgrind's fault.  If you think this is the case or
==3862==    you are not sure, please let us know and we'll try to fix it.
==3862== Either way, Valgrind will now raise a SIGILL signal which will
==3862== probably kill your program.
```

I did not see a warning about a bad jump, so I assume this is Valgrind's fault. It would be nice to know which version of everything we're using here, so we may fix the problem.

cfogelklou: Valgrind on github actions is using 3.18. I am guessing your local machine is on 3.15. However, I think I have found a fix - installing valgrind via homebrew.

LoupVaillant: Or, GCC is more recent on Github, and generates instructions Valgrind doesn't handle yet. Here's what I get on my machine:

```
$ uname -a
Linux old-grey 5.13.0-28-generic #31-Ubuntu SMP Thu Jan 13 17:41:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

$ gcc --version
gcc (Ubuntu 11.2.0-7ubuntu2) 11.2.0

$ valgrind --version
valgrind-3.17.0
```

If you have a quick fix, it would be nice if you could apply it in a third commit of your pull request. If it doesn't work that's too bad, but what we have now is still worth merging.

cfogelklou: My "homebrew fix" was very flakey. It worked once, then didn't work again (even rerunning the same commit.)
Still checking for a solution...

cfogelklou: I did "resolve" in this action: https://github.com/cfogelklou/Monocypher/actions/runs/1955574996
However, the fix was to build with CMake & Clang instead of GCC.
That same merge also adds cross-platform building testing on windows and macOS :)

LoupVaillant: Hmm, There’s probably a way to build with Clang and regular Make, then run Valgrind on that. We should test that.

Just one more thing before I merge: is there a way to add comments at the beginning of the GitHub actions file? If so, it would be nice if you could add the licence info like there is in pretty much all other files (just copy one at random, and adjust the author name & year).

cfogelklou: Sure. I'll add the comments and try building with Clang. I'll do that and then update this PR. Hold tight.

fscoto: Maybe I'm thinking too far ahead, but maybe there's a weird interaction between valgrind and gcc's codegen in those particular versions.

What about replacing `ubuntu-latest` with some other kind of Linux base image?

(Evidently, all other tests run normally, so it's not like there's a direct codegen bug in gcc itself, since the normal tests with illegal instructions should fail equally.)

cfogelklou: Hi, I found the issue. It doesn't matter if clang or gcc is used, the culprit is the flag "-march=native", regardless of optimization level.

I think it is a good idea to leave failing tests failing, so the failing tests are left in, but ignored. But the workflow finishes with passing valgrind runs, so any subsequent failures introduced by anyone will cause those tests to fail.

You can check the result here: 
[https://github.com/cfogelklou/Monocypher/runs/5506724695?check_suite_focus=true
](https://github.com/cfogelklou/Monocypher/runs/5506724695?check_suite_focus=true
)

LoupVaillant: Looks good to me, merging. Thanks.

If the only culprit is `march=native`, I think we can safely ignore the problem, especially if we have establish that it's Valgrind's fault for not handling the instruction — the program works after all.

PR #239, opened 2022-06-16
makefile: fix cross-compilation

sikmir:
https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html:

> AR -  Archive-maintaining program; default ‘ar’.


LoupVaillant:
Thanks for the quick fix! Works on my machine, merging now.

PR #249, opened 2023-01-08
Simplify crypto_verify

luizsdf:
Allows `crypto_verify` to work with buffers of arbitrary size.

### DIFF
--- a/doc/man/man3/advanced/crypto_poly1305.3monocypher
+++ b/doc/man/man3/advanced/crypto_poly1305.3monocypher
@@ -122,7 +122,7 @@ in bytes.
 produces a message authentication code for the given message and
 authentication key.
 To verify the integrity of a message, use
-.Xr crypto_verify16 3monocypher
+.Xr crypto_verify 3monocypher
 to compare the received MAC to the output
 .Fa mac .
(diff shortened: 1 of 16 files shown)


LoupVaillant:
Oh, my, I'm so sorry.

I applaud the attempt, but what you just did was reverting to the approach I attempted in the first version of Monocypher, and shied away from because the generated code was convoluted as hell, to the point where we had reason to suspect it introduced variable-time stuff even though the C code was naturally constant time. I did this for version 1.1.0, and the changelogs sadly failed to reflect that. I shall correct this oversight ASAP.

Another advantage of the fixed size buffer comparisons is that the generated code is extremely small and fast. The compiler compares whole _words_ at a time, generally in unrolled loops. The typical 64-byte buffer comparison in a 64-bit machine would involve just 8 reads per buffer (16 total), which is both extremely fast and very compact.

Long story short, I believe the fixed sized buffer comparisons are here to stay. I agree it sucks, having a single function would be better. But I'm not touching that with a 10-foot pole until we can actually trust the generated code.

luizsdf:
Wow, sorry, I really had no idea!

PR #251, opened 2023-01-18
Fix path of header containing getrandom

noocsharp:


LoupVaillant:
Looks good to me, thanks for the fix!

PR #252, opened 2023-01-20
Document new structure-based APIs

fscoto:


doc/man/man3/crypto_argon2.3monocypher
```diff
@@ -132,30 +111,92 @@ If you are not sure how to allocate that buffer, just use
...
 It must be at least 1.
 A value of 3 is
 .Em strongly
-recommended;
+recommended when using Argon2i;
 any value lower than 3 enables significantly more efficient attacks.
-.It Fa password
+.It Fa nb_lanes
+The level of parallelism.
+This parameter has no effect as Monocypher has no concept of threading.
```
LoupVaillant, line 173: Strictly speaking, this parameter *does* have an effect, since the resulting hash will conform to what the Argon2 specification says. It's just that:

- Even though it _could_ be computed in parallel with at most `nb_lanes` threads, Monocypher uses only 1 thread.
- Therefore, using any value other than 1 has the very real effect of increasing attacker advantage, effectively _reducing_ security.

By the way, I recall having read a paper suggesting that users who want actual parallelism should instead run several instances of Argon2 in parallel themselves, (using different additional data for each thread), then hash the results together in one final hash.

doc/man/man3/crypto_argon2.3monocypher
```diff
@@ -245,6 +264,40 @@ Length of
...
+Since parameter selection depends on your hardware, some trial and error
+will be required in order to determine the ideal settings.
+Three iterations and 100000 blocks
+(one hundred megabytes of memory)
+is a good starting point.
+Adjust
+.Fa nb_blocks
+first.
+If using all available memory is not slow enough, increase
+.Fa nb_passes .
```
LoupVaillant, line 300: We may want to revise this section based on current recommendation. I'm thinking of scoobz's advice Soatok outlined in his recent article.

Or we may not. I believe this procedure is still good. The starting point may need to be increased up to 300MB. Perhaps. I'm not sure.

LoupVaillant (commented):
A quick first pass suggest your work is outstanding, thanks a lot.

I have just one nitpick where you say `nb_lanes` has no effect that may be a bit confusing. If you have an idea how you want to change it go ahead, or I can merge and I'll think of a reformulation.

Separately, we may want to add a section somewhere about the recommended way to achieve parallelism (and I need to trace the source of this recommendation). If you know where to put that do it, or we can merge this now and add it later. Same thing about the old procedure to chose parameters, we may want to adjust them a bit with the new state of the art, and the fact that we now support Argon2id and Argon2d.

I'll do a less quick pass tomorrow, but so far this already looks good to merge.

LoupVaillant:
Better merge it now, we can sort out the details later. Thanks again.

doc/man/man3/crypto_argon2.3monocypher
samuel-lucas6: What makes you say 300MB? Sc00bz's [website](https://tobtu.com/minimum-password-settings/) allows for much lower as a baseline.

LoupVaillant: I was just throwing that around. I ended up not changing my original numbers. (Well I _did_ add an Argon2id baseline that's 300MB, but that's because we typically use it with only one pass, while Argon2i needs 3).

Now ScOObz is describing _minimum_ settings, and it's not clear for which use case. A login in a low-enough stakes website is certainly cool with that. Your password manager's master password though? probably needs more.

Anyway, apparently I was panicking for no reason, thanks for the link. I can resolve the thread now.

doc/man/man3/crypto_argon2.3monocypher
LoupVaillant: I have now modified it to my satisfaction. Further corrections are welcome if I messed it up though.

Issue #253, opened 2023-01-21
Support Blake2s for embedded targets


LoupVaillant:
I've been approached lately by @Laczen about possibly making an "embedded edition" of Monocypher, less focused on speed and more focused on reducing footprint (binary & stack sizes). One particular step in this direction is implementing Blake2s.

I have resisted Blake2s for a long time, most notably because of the redundancy with Blake2b (we already have SHA-512 for compatibility, I don't want to make things even worse), but I have to admit it makes a whole lot of sense for what is arguably our biggest user base.

Ideally we would be able to share code between Blake2b and Blake2s, and have Blake2s be supported by a flag with the current Blake2 functions. To be checked, but I believe such sharing is not actually possible because the internal loops of Blake2b use 64-bit integer, while Blake2s uses 32-bit integers instead. So we should probably provide Blake2s separately, in a new optional compilation unit: `src/optional/blake2s.h` and `src/optional/blake2s.c`.

There's also Blake3 to consider, since it's 32-bit too, but takes advantage of vector instructions for bigger processors. I have two problems with it though: first, it has tighter security margins, in accordance to Aumasson's [Too Much Crypto](https://eprint.iacr.org/2019/1492.pdf). Not exactly unsafe, but perhaps a bit risky? Second, it would replace Blake2b, which would be a hard break, and so far that never happened with Monocypher (we broke the API several times, but we never ever completely broke the ability of users to keep their old wire formats —even for AEAD they could use the basic primitives to achieve it again). Another problem with removing Blake2b would be the inability to implement Argon2. A variant would be easy for sure, but we'd lose the standard compliance. Oh, and I recall @fscoto warning me that Blake3 needs a bigger footprint than Blake2s, making it is less ideal for small embedded targets.

Anyway, adding Blake2s would be a step towards the support of embedded targets.

The next step would be writing an actual embedded edition of Monocypher. Something that sacrifices speed so it can be smaller (make sense with the smallest targets, that have limited program memory, and teeny tiny stacks). The problem here is how far we should be willing to go, and that very much depends on the embedded market: are 8-bit and 16-bit processors relevant? Would we even try to do cryptography on them? How about extremely low-powered 32-bit processors? Do they have enough memory so we should optimise for speed to reduce energy consumption? Or do they need smaller crypto code? What multipliers should our bignum arithmetic target? Are the current 32->64 bit multiplications okay, or should we go down to 16->32, or even 8->16 like C25519 does?

[...]

PR #254, opened 2023-01-21
Added X25519_NO_UNROLLING option

Laczen:
Added X25519_NO_UNROLLING option to reduce binary size. This reduces speed but allows a user to trade in some performance for size.

Signed-off-by: Laczen JMS <laczenjms@gmail.com>


src/monocypher.c
```diff
@@ -1178,6 +1235,29 @@ static void fe_mul_small(fe h, const fe f, i32 g)
...
+
+    	FOR (i, 0, 10) {
+        	t[i] = 0;
+		FOR (j, 0, i + 1) {
+            		if (((i & 0x1) == 0) && ((j & 0x1) == 1)) {
+                		t[i] += f[j] * (i64)(g[i - j]) * 2;
+            		} else {
+                		t[i] += f[j] * (i64)(g[i - j]);
+            		}
+        	}
```
LoupVaillant, line 1249: I have the feeling here that we can be a bit smarter about the test, perhaps avoid it entirely. This would likely involve a tiny bit of unrolling, but we would likely get some performance back.

Also, if we're okay with wasting a bit of stack, we should be able to pre-compute the multiplications by 2 (and below, by 19 and 38). While the factor of 2 may not be worth it (it's a simple shift by 1, very fast everywhere), pre-computing the multiplication by 19 is likely to recover quite a bit of speed.

I need to experiment some with this.

LoupVaillant: Interesting. I would have thought that having 2 inner loops (from 0 to i and from i to 10) would have been a bit faster. Good job there.

LoupVaillant: Okay, I believe I have found something even better:

```C
	lfe t;
	fe g19;
	FOR (i, 0, 10) {
		g19[10 - i] = g[i] * 19;
	}

	FOR (i, 0, 10) {
		t[i] = 0;
		FOR (j, 0, 10) {
			i64 prod = f[j] * (i64)(j > i
			                        ? (g19[j - i])
			                        : (g[i - j]));
			prod <<= ~i & j & 0x1;
			t[i] += prod;
		}
	}

	fe_carry(h, t);
```

The idea was hoisting out the multiplication by 19, which I thought cost quite a bit. The main problem doing that is that we're sacrificing stack size (it's probably 40 bytes bigger than your proposal). I also took the opportunity to simplify your test, and I found this neat arithmetic trick to achieve what you did. Anyway, my change here ultimately makes X25519 quite a bit faster on my laptop:

- 7% faster than your code if we compile under `-O3 -march=native`.
- 16% faster than your code if we compile under `-Os -march=native`.

Now one thing bothers me about this whole patch (whatever the version): buffers aren't wiped. See, all arrays that may contain secret data are wiped before they're freed. It's kind of a "best effort" thing, most notably because ordinary local variables, more likely to live in registers, aren't wiped at all. And in the case of X25519 field multiplications some of those local variables _definitely_ spill on the stack, and they're not wiped either. Still, I wanted to note that those two little arrays we just added? They would be an exception to the "wipe all the buffers!" rule, and I don't like that.

While I know it's not entirely rational, I now feel compelled to look at this a bit more critically. Let's measure this all under various conditions. I'll be using GCC 11.3.0 on Ubuntu 22.04, on my x86-64 Intel Skylake laptop. Let's start with the sizes of `monocypher.o`, in bytes (each column compares itself to the previous column):

| | Unrolled | use fe_mul() | unroll (me) | unroll (you) |
| --- | --- | --- | --- | --- |
| __-Os march=native__ | 53,720 | -1232 | -2160 | -32 |
| __-O2 march=native__ | 57,552 | -1384 | -2168 | -24 |
| __-O3 march=native__ | 86,064 | -1216 | -592 | +296 (!!) |

And now the speed, in number of exchanges per second:

| | Unrolled | use fe_mul() | unroll (me) | unroll (you) |
| --- | --- | --- | --- | --- |
| __-Os march=native__ | 7050 | 6052 (-14%) | 2099 (-66%) | 1804 (-14%) |
| __-O2 march=native__ | 7402 | 6393 (-14%) |1889 (-70%) |1656 (-12%) |
| __-O3 march=native__ | 7641 | 6494 (-15%) | 5458 (-16%) | 5064 (-7%) |

First, the obvious: someone who needs your patch would _never_ use `-O3`, because the binary is just much bigger. Still, I'm impressed by how well performance is holding up under `-O3`: the smallest binary (mine) manages to keep total performance drop  under 29%, much less than what I originally measured.

Another unsurprising finding is the effect of replacing `fe_sq()` by a single call to `fe_mul()`. This change alone saves between 1.2 and 1.3 kilobytes, and the performance drop-off is very stable.

More notable is the resilience of the unrolled code code to compilation options: going from `-O3` to `-Os` only costs us 8% speed (the `use fe_mul()` version is just as resilient with a 7% drop in speed).

The real problems however start when we're running the fully rolled code under `-O2` or `-Os`. The performance drops are _massive_ here. The most generous comparison I can make (my version compared to the `fe_mul()` only version) still shows that we're dividing speed by more than 3! All this to save less than 2.2KB on the binary. I know embedded devices are tight on memory, but the smallest ones tend to be tight on _speed_ as well. Saving 2.2KB is good, but a 3x slowdown is no joke.

My tentative conclusions right now:

- Using `fe_mul()` instead of `fe_sq()` saves ~1.3KB, and cost less than 15% of speed. Definitely worth it in some circumstances. Not sure I want to accept a patch just for this though, because it's so easy to do it yourself. What I _can_ do is put a nice comment that explains how to do it and what to expect exactly. Given the change, and my observations, I believe the effects should be stable across platforms as well.

- Rolling everything like you did however, even with my performance improvements, is likely going too far: saving 2.2KB is nice, but I think dividing performance by more than 3 will be too expensive in most cases. It's a "desperate measures for a desperate cases" kind of deal, too niche in my opinion to accept the full patch right now.

One caveat in my analysis: I haven't analysed the impact of only inlining carry propagation. I expect a fairly stable performance impact, but I don't think it will save much more than 800 bytes of code: after we've simplified `fe_sq()` the carry routine is used only 3 times, and the carry propagation logic is not that big.

---

Sorry for giving you this bummer, but I must say on the other hand reviewing this patch has been a pleasure. Even though I'm now likely to reject it it was interesting and valuable research, that I will keep with me going forward when I do get around to make that embedded edition for real.

Laczen: @LoupVaillant, thank you for taking the time to study this.

On my ryzen7 4700u I see a small reduction in speed with your proposal.

Regarding the size, I am focussing on the size of the X25519 part (as this is what is being changed). I am getting the following results (all compiled with -Os):
nrf51422 (cortex m0): from 8070kB reduced to 2380kB,
nrf52832 (cortex m4): from 5036kB reduced to 2444kB,
stm32f103 (cortex m3): from 5048kB reduced to 2456kB,

This is a reduction of more than 50%.

Regarding the wiping of buffers, there should be no exception. However since the t variable is created only for passing it to `fe_carry` it is very easy to set them to zero when copying to `h` at the end of `fe_carry`.

Laczen: And also, don't forget the benchmarking figures presented by Zandberg et al, monocypher was so much faster that a increase of even a factor 3 should still place monocypher on top at possibly comparable binary and stack usage.

LoupVaillant: > On my ryzen7 4700u I see a small reduction in speed with your proposal.

Okay… go figure I guess? In any case that's not very important: the real target here are the Cortex cores. And for those… _Damn_. I have to admit that cutting code size in half can indeed justify a 3-fold slowdown. But at this point I won't even speculate at the actual effect. I still remember the `BLAKE2B_NO_UNROLLING` option _increasing_ performance in some target (third party report, but they seemed trustworthy).

Anyway, let's compare our numbers here:

| | Unrolled | Rolled | bytes saved | % saved |
| --- |--- |--- |--- |--- |
| __x86-64__ (me) | 9904 | 6528 | 3376 | 34% |
| __x86-64__ (you) | 9904 | 6496 | 3408 | 34% |
| __Cortex M4__ | 5036 | 2444 | 2592 | 51% |
| __Cortex M3__ | 5048 | 2456 | 2592 | 51% |
| __Cortex M0__ | 8070 | 2380 | 5690 | 71% |

_(I believe the numbers you gave were meant to be bytes.)_

Okay, so the biggest binary is the x86-64 one. Could be some overhead from Linux, but considering the savings on the M3 and M4 I would suspect the denser ARM instruction set first. Then quite unsurprisingly the savings on the M3 and M4 are absolutely identical. The outlier here really is the M0. It manages to get a much bigger binary before your unrolling, and a slightly smaller one _after_. The effects of rolling here are _massive_. And let's be honest, if there's one 32-bit chip that may be short on memory, it's this one.

Clearly, the savings on the M0 totally justify dividing the speed by three in some cases (but again, until we measure the actual performance impact is anyone's guess). At least I now know we should give users the option one way or another.

More important though is how the unrolled code sizes suggest the M0 lacks a relevant instruction. Something like mul-add, perhaps even a 32->64 bit multiply instruction. If it's the latter, we _need_ that embedded edition. And that won't be a mere rolling of the current code, the size of the limbs will have to change as well. This will require a complete overhaul of the bignum arithmetic, and with any luck we'll be able to improve size _and_ speed on the M0.

I need to read the M0 data sheet, and get my hand on a cycle accurate simulator, or a real chip.

This pull request is likely to remain open for a long time. My hope is to subsume it with an embedded edition, but until this is done I'd like to keep it open as a reminder.

LoupVaillant: Digging a little deeper, I can see that the various cortices use slightly different instruction sets:

- __Cortex M0:__ armv6-M
- __Cortex M3:__ armv7-M
- __Cortex M4:__ armv7E-M

This [link](https://community.arm.com/arm-community-blogs/b/architectures-and-processors-blog/posts/armv6-m-vs-armv7-m---unpacking-the-microcontrollers) is saying that armv6-M is a _simplification_ of armv7-M, and the corresponding cortices are even more recent than the M3.

And [there you have it](https://usermanual.wiki/Pdf/DDI0419Carmarchitecturev6mreferencemanual.820733959/view), section A4.4.3:

> The only multiply instruction supported in ARMv6-M performs a 32x32 multiply that generates a 32-bit 
result, see _MUL_ on page A6-159. The instruction can operate on signed or unsigned quantities.

This almost certainly means bignum arithmetic needs to avoid 64-bit integers. There's our justification for the 32-bit embedded edition.

From [here](https://developer.arm.com/documentation/ddi0403/d/Application-Level-Architecture/The-ARMv7-M-Instruction-Set/Data-processing-instructions/Multiply-instructions?lang=en) however we can see that the M3 and M4 have access to the juicy 32->64 bit _multiply and accumulate_ instruction that makes Monocypher so fast on those targets. This also explains the smaller binary sizes: with the mul-add instructions we can generate smaller bignum code.

Also, a quick look at the [Wikipedia](https://en.wikipedia.org/wiki/STM32#Part_number_decoding) suggests that the M3 and M4 (in the STM32 chips) have much more available memory than the M0 and M0+. This seems to confirm my intuition that we should focus our efforts on an embedded edition addressing the M0 and similar chips that lack the 32->64 bit multiplier, and let the more powerful chips use the main edition. Worst case, users will still be able to use the embedded edition to get the smaller binary.

LoupVaillant: (Summoning @fscoto who is sure to have a relevant opinion about this, and tagging #253 where this all started.)

Laczen: @LoupVaillant, I will do some performance benchmarks on real silicon to get an idea of performance.

Laczen: I have taken some time to do some trials on real silicon (nucleo_f411re) and qemu. The times given are for one X25519 exchange:
| | unrolled | rolled (my version, added t[] wipe)|ratio|
|-|----------|---------------------|-----|
|nucleo_f411re (84MHz)|20ms|96ms|x4.8|
|cortex_m3 (qemu 12MHz)| 82ms|418ms|x5.1|
|cortext_m0 (qemu 1MHz)| 1216ms| 2513ms|x2.1|

This leaves me with mixed feelings: on cortex m3/m4 there is a code reduction of about a factor 2, but a speed decrease of a factor 5. On cortex m0 there is a code reduction of 3.5 with a speed decrease of a factor 2.

LoupVaillant: Thanks for those trials, those are really informative!

If we were serious about the wipe, we would hoist the array out of the function, and wipe it only once.
Kind of messy now that every single call to `fe_mul()` comes with an extra argument, but at least you only wipe once. Should help a bit.

Could you test this quick _without_ the wipe so we get a feeling of the performance we might hope for?

Laczen: The influence is neglectable (4ms on qemu cortex_m3), the way t is cleared is by setting it to 0 in the loop that converts t to h. t was only created to allow passing it to carry_fe(h,t).

Laczen: BTW, regarding the figures for cortex m3/m4 this might not be so bad after all given the speed comparison (Zandberg et al): monocypher was 40 times faster than C25519.

Issue #264, opened 2023-10-19
Code in manual has 1 error + 1 warning - requires changes that could result in unexpected results


SethArchambault:
I'm trying to get a very basic use of argon2 working, but not matter what I do it won't match the correct results. I'm sure it's something simple, but I'm finding it very hard to debug.

Salt: ffffddddffffdddd
Password: OkayPassword 

```
Output In Hex Form:
68e427376e16216f19425333ccdcc688d51b92d31b7ba1b2096b6f450254e850
Output in Encoded Form:
$argon2i$v=19$m=4096,t=3,p=1$ZmZmZmRkZGRmZmZmZGRkZA$aOQnN24WIW8ZQlMzzNzGiNUbktMbe6GyCWtvRQJU6FA
```

When I compare this to running the hashing algorithm from this website I get:
https://argon2.online/
```
Output In Hex Form:
02f9d1e398c0db9676e44550c03e77a4a0d93c74e90c5966d3dd48756a65c5bd
Output in Encoded Form:
$argon2i$v=19$m=4096,t=3,p=1$ZmZmZmRkZGRmZmZmZGRkZA$AvnR45jA25Z25EVQwD53pKDZPHTpDFlm091IdWplxb0
```

<img width="1172" alt="image" src="https://github.com/LoupVaillant/Monocypher/assets/1406558/886059dc-cc88-44ab-be4a-0aa5f3c4541f">

And the same results show in command line:
```
$ echo -n OkayPassword | argon2 ffffddddffffdddd -i -t 3 -k 4096 -p 1
Type:           Argon2i
Iterations:     3
Memory:         4096 KiB
Parallelism:    1
Hash:           02f9d1e398c0db9676e44550c03e77a4a0d93c74e90c5966d3dd48756a65c5bd
Encoded:        $argon2i$v=19$m=4096,t=3,p=1$ZmZmZmRkZGRmZmZmZGRkZA$AvnR45jA25Z25EVQwD53pKDZPHTpDFlm091IdWplxb0
0.019 seconds
Verification ok
```

Here's my code:

build.sh
```
set -e
clang++ -g main.cpp
./a.out
```

main.cpp
```
#include <stdio.h>
#include <stdlib.h>
#include "monocypher.h"
#include "monocypher.c"

#define assert(expr) if(!(expr)) { printf("%s:%d %s() %s\n",__FILE__,__LINE__, __func__, #expr); *(volatile int *)0 = 0; }

#include <stdint.h>
[...]

PR #265, opened 2023-10-19
Update crypto_argon2.3monocypher - fixes errors and warnings that can lead to the incorrect solution

SethArchambault:
Closes: #264 

Fixes these issues, by allowing password array to autosize, and then making sure to drop the \0 character when determining the size of the string, and also reorders the crypto_argon2_inputs field designators to remove a warning.

```
main.cpp:83:24: error: initializer-string for char array is too long, array size is 14 but initializer has size 15 (including the null terminating character)
uint8_t password[14] = "Okay Password!";
                       ^~~~~~~~~~~~~~~~
main.cpp:87:5: warning: ISO C++ requires field designators to be specified in declaration order; field 'pass_size' will be initialized after field 'salt' [-Wreorder-init-list]
    .salt      = salt,                 /* Salt for the password */
    ^~~~~~~~~~~~~~~~~
```

### DIFF
--- a/doc/crypto_argon2.3monocypher
+++ b/doc/crypto_argon2.3monocypher
@@ -341,11 +341,11 @@ crypto_argon2_config config = {
     .nb_passes = 3,                          /* 3 iterations    */
     .nb_lanes  = 1                           /* Single-threaded */
 };
-uint8_t password[14] = "Okay Password!";
+uint8_t password[] = "Okay Password!";
 crypto_argon2_inputs inputs = {
-    .pass      = password,                     /* User password */
-    .pass_size = sizeof(password),           /* Password length */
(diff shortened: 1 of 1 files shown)


LoupVaillant:
Looks pretty good to me, thanks for the fix!

This may help more than you realise, such issues not only waste time (sorry about yours), they also tend to hurt whatever trust users may have in the library.

SethArchambault:
@LoupVaillant Cool, glad it helps! 😄 

Appreciate your work, and glad this exists! 👍

SethArchambault:
Realizing now that I was conflating strlen with sizeof! 🤦  strlen would be a fine replacement for sizeof here that would avoid all confusion.

PR #266, opened 2023-10-20
Had to change soname to install_name on macos, but I don't really know why

SethArchambault:
Was getting this error:
```
$ make
cc -pedantic -Wall -Wextra -O3 -march=native -L/opt/homebrew/opt/llvm/lib -shared -Wl,-soname,libmonocypher.so.4 -o lib/libmonocypher.so.4 lib/monocypher.o lib/monocypher-ed25519.o
ld: unknown options: -soname
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [lib/libmonocypher.so.4] Error 1
```

And someone on stackoverflow said to change soname to install_name, so I did that:
https://stackoverflow.com/questions/4580789/ld-unknown-option-soname-on-os-x

And it worked! But not very scientific and could have downsides.. But figured I'd let ya know.

-	$(CC) $(CFLAGS) $(LDFLAGS) -shared -Wl,-soname,$(SONAME) -o $@ $(MAIN_O)
+	$(CC) $(CFLAGS) $(LDFLAGS) -shared -Wl,-install_name,$(SONAME) -o $@ $(MAIN_O)

### DIFF
--- a/makefile
+++ b/makefile
@@ -132,7 +132,7 @@ lib/libmonocypher.so: lib/$(SONAME)
 	ln -sf `basename lib/$(SONAME)` $@
 
 lib/$(SONAME): $(MAIN_O)
-	$(CC) $(CFLAGS) $(LDFLAGS) -shared -Wl,-soname,$(SONAME) -o $@ $(MAIN_O)
+	$(CC) $(CFLAGS) $(LDFLAGS) -shared -Wl,-install_name,$(SONAME) -o $@ $(MAIN_O)
 
 lib/monocypher.o: src/monocypher.c src/monocypher.h
 	@mkdir -p $(@D)


LoupVaillant:
Hmm, good catch, but I’ll need to investigate. I have a couple fixes in mind:

1. Accepting your change, if it turns out it works with other compilers and linkers.
2. Put the `-soname` option itself into a variable instead of hard coding it.
3. Use `ld` or `$(LD)` or similar instead of `$(CC)` for the link step. May be combined with step 2 to work with clang, should someone use it to, say, enable link-time-optimisation or use a special link script.

Putting it on hold for now, I should get to it in one or two weeks. In the mean time, if you fancy doing some testing let me know what you find, this’ll help me.

aikawayataro:
> but I don't really know why

No soname for Darwin, but there's a similar feature "Install name". It was used by you as a workaround but it's Darwin-only, so no way this gonna work anywhere else.

PR #275, opened 2024-05-18
Doc: use pointers instead of arrays when size is provided

noocsharp:


LoupVaillant:
My, I should have done that myself ages ago. Thanks, merged.

PR #276, opened 2024-06-20
Fix problamatic line

sagarreddypatil:
This line of code looked like an out of bounds access within the context of the function.
However, it's a static function and the size of secret_key at all the caller locations was 64


LoupVaillant:
Oops, good catch, thanks for the fix! Merging now.

PR #290, opened 2026-07-27
headers: Match C++ namespace guard opening and closing conditions

rgetz:
The public headers conditionally open either a C++ namespace or an `extern "C"` block:
```
    #ifdef MONOCYPHER_CPP_NAMESPACE
    namespace MONOCYPHER_CPP_NAMESPACE {
    #elif defined(__cplusplus)
    extern "C" {
    #endif
```
However, the closing preprocessor condition only tested `__cplusplus`, rather than mirroring the opening logic.

While this did not affect normal C++ compilation (where `MONOCYPHER_CPP_NAMESPACE` implies `__cplusplus`), it left the header logic asymmetric and could (and does) confuse some static analysis tools (such as cppcheck), resulting in syntax errors during parsing.

Update the closing guards to check `MONOCYPHER_CPP_NAMESPACE` first, followed by `__cplusplus`, matching the opening conditions and making the conditional structure consistent and self-contained.


LoupVaillant:
Good catch, thanks. Merging right away.

